PR #39, opened 2016-12-13
Using the new auth endpoint for SpeechToText

iassal:


iassal:
@willportnoy @kgrashad @ritika2502

PR #86, opened 2017-03-09
[CSharp] DirectLine + WebSockets sample

ejadib:


CSharp/core-DirectLineWebSockets/DirectLineClient/Program.cs
```diff
@@ -0,0 +1,118 @@
...
+    using Newtonsoft.Json;
+    using System.Threading;
+    using System.Text;
+    using WebSocketSharp;
+
+    public class Program
+    {
+        private static string directLineSecret = ConfigurationManager.AppSettings["DirectLineSecret"];
+        private static string botId = ConfigurationManager.AppSettings["BotId"];
+        private static string fromUser = "DirectLineSampleClientUser";
```
dandriscoll: Can you add the following comment above fromUser?

// fromUser is the field that identifies which user is sending activities to the Direct Line service.
// Because this value is created and sent within your Direct Line client, your bot should not
// trust the value for any security-sensitive operations. Instead, have the user log in and
// store any sign-in tokens against the Conversation or Private state fields. Those fields
// are secured by the conversation ID, which is protected with a signature.

CSharp/core-DirectLineWebSockets/DirectLineClient/Program.cs
```diff
@@ -0,0 +1,118 @@
...
+                            await directLineClient.Conversations.PostActivityAsync(conversation.ConversationId, userMessage);
+                        }
+                    }
+                }
+            }
+        }
+
+        private static void WebSocketClient_OnMessage(object sender, MessageEventArgs e)
+        {
+            // avoid null reference exception when no data received
```
dandriscoll: Change this comment to:

// Occasionally, the Direct Line service sends an empty message as a liveness ping. Ignore these messages.

CSharp/core-DirectLineWebSockets/DirectLineClient/packages.config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Microsoft.Bot.Connector.DirectLine" version="3.0.0" targetFramework="net461" />
+  <package id="Microsoft.Rest.ClientRuntime" version="2.3.4" targetFramework="net461" />
```
dandriscoll: We've hit binary incompatibilities between minor versions of Microsoft.Rest.ClientRuntime. 2.3.2 is probably a better example to use.

CSharp/core-DirectLineWebSockets/README.md
```diff
@@ -0,0 +1,136 @@
...
+````C#
+private static void WebSocketClient_OnMessage(object sender, MessageEventArgs e)
+{
+    var activitySet = JsonConvert.DeserializeObject<ActivitySet>(e.Data);
+    var activities = from x in activitySet.Activities
+        where x.From.Id == botId
+        select x;
+````
+
+DirectLine v3.0 (unlike version 1.1) has support for Attachments (see [Adding Attachments to a Message](https://docs.botframework.com/en-us/core-concepts/attachments) for more information about attachments). Check out the `WebSocketClient_OnMessage` method in [Program.cs](DirectLineClient/Program.cs#L73-L90) to see how the Attachments are retrieved and rendered appropriately based on their type.
```
dandriscoll: In printed text, Direct Line is normally two words.

It's OK to collapse them in filenames, though. ("DirectLineClient.cs" is OK)

CSharp/core-DirectLineWebSockets/DirectLineClient/Program.cs
dandriscoll: One other item I just thought of: can you add some code to the sample that exchanges the Direct Line secret for a token before using it?

CSharp/core-DirectLineWebSockets/DirectLineClient/Program.cs
pcostantini: [Done](https://github.com/Microsoft/BotBuilder-Samples/pull/86/files#diff-a8f081a979c6ac90c9b08ab44d92b01fR19)

CSharp/core-DirectLineWebSockets/DirectLineClient/Program.cs
pcostantini: Done and [added call to Tokens API to retrieve token](https://github.com/Microsoft/BotBuilder-Samples/pull/86/files#diff-a8f081a979c6ac90c9b08ab44d92b01fR34), also added [note to README](https://github.com/Microsoft/BotBuilder-Samples/pull/86/files#diff-88e83e4a6e98da048a5d48e8dc9836d1R33)

CSharp/core-DirectLineWebSockets/DirectLineClient/packages.config
pcostantini: Done

CSharp/core-DirectLineWebSockets/README.md
pcostantini: Fixed in Node too

pcostantini:
Hi @dandriscoll , we've updated the PR with the latest feedback.
Regards!

PR #87, opened 2017-03-09
[Node] DirectLine + WebSockets sample

ejadib:


Node/core-DirectLineWebSockets/DirectLineClient/app.js
```diff
@@ -0,0 +1,195 @@
+var Swagger = require('swagger-client');
+var open = require('open');
+var rp = require('request-promise');
+
+// Config settings
+var useW3CWebSocket = false;
+var directLineSecret = 'DIRECTLINE_SECRET';
+var directLineClientName = 'DirectLineClient';
```
dandriscoll: directLineUserId may be better

dandriscoll: Can you add the following comment above this variable?

// directLineUserId is the field that identifies which user is sending activities to the Direct Line service.
// Because this value is created and sent within your Direct Line client, your bot should not
// trust the value for any security-sensitive operations. Instead, have the user log in and
// store any sign-in tokens against the Conversation or Private state fields. Those fields
// are secured by the conversation ID, which is protected with a signature.

Node/core-DirectLineWebSockets/DirectLineClient/app.js
```diff
@@ -0,0 +1,195 @@
...
+var directLineSpecUrl = 'https://docs.botframework.com/en-us/restapi/directline3/swagger.json';
+
+process.argv.forEach(function (val, index, array) {
+  if (val === 'w3c') {
```
dandriscoll: Can we just call the option "websocket" ? I don't think we refer to Web Sockets as W3C Web Sockets anywhere else. They're just regular Web Sockets to us :)

Node/core-DirectLineWebSockets/DirectLineClient/app.js
```diff
@@ -0,0 +1,195 @@
...
+
+var directLineClient = rp(directLineSpecUrl)
+    .then(function (spec) {
+        // Client
+        return new Swagger({
+            spec: JSON.parse(spec.trim()),
+            usePromise: true
+        });
+    })
+    .then(function (client) {
```
dandriscoll: Can you also illustrate using the Token API to exchange the client's secret for a token?

Node/core-DirectLineWebSockets/DirectLineClient/app.js
dandriscoll: Actually, I noticed below that W3C is a specific kind of socket exposed by Node? If so, this is OK.

Node/core-DirectLineWebSockets/DirectLineClient/app.js
```diff
@@ -0,0 +1,195 @@
...
+    ws.on('connect', function(connection) {
+        console.log('WebSocket Client Connected');
+        connection.on('error', function(error) {
+            console.log("Connection Error: " + error.toString());
+        });
+        connection.on('close', function() {
+            console.log('WebSocket Client Disconnected');
+        });
+        connection.on('message', function(message) {
+            if (message.type === 'utf8' && message.utf8Data.length > 0) {
```
dandriscoll: Comment similar to:

// Occasionally, the Direct Line service sends an empty message as a liveness ping. Ignore these messages.

Node/core-DirectLineWebSockets/README.md
```diff
@@ -0,0 +1,181 @@
+# Direct Line Bot Sample (using client WebSockets)
+
+A sample bot and a custom client communicating to each other using the Direct Line API.
```
dandriscoll: Somewhere near the top of the sample, we should note the following:

* For many developers, the [DirectLineJs ](https://github.com/Microsoft/BotFramework-DirectLineJs) project is an easy way to connect to the Direct Line in Node.js
* If you're interested in writing your own client without using DirectLineJs, use this sample

Node/core-DirectLineWebSockets/DirectLineClient/app.js
pcostantini: It is the same library, but the API exposed by the two versions change. The NodeJS version uses EventEmitter to trigger events, while the 'w3cwebsocket' follows the W3C spec using event handlers set as props, which is what WebBrowsers do.

Node/core-DirectLineWebSockets/DirectLineClient/app.js
pcostantini: [Done](https://github.com/Microsoft/BotBuilder-Samples/pull/87/files#diff-5b5377952ce775d849ca57207e03cdb8R8)

Node/core-DirectLineWebSockets/DirectLineClient/app.js
pcostantini: [Done](https://github.com/Microsoft/BotBuilder-Samples/pull/87/files#diff-5b5377952ce775d849ca57207e03cdb8R37). Also added note on README [here](https://github.com/Microsoft/BotBuilder-Samples/pull/87/files#diff-dd1c1c9e6d8e175147fa06e11004141fR35).

Node/core-DirectLineWebSockets/README.md
pcostantini: [Done](https://github.com/Microsoft/BotBuilder-Samples/pull/87/files#diff-dd1c1c9e6d8e175147fa06e11004141fR5)

pcostantini:
@dandriscoll Hi Dan, the feedback for the Node's DirectLine with WebSockets has been applied.
Thanks!

PR #100, opened 2017-04-02
updating LUIS model URL

v-youel:
changing LUIS model url to new one because the only one will be deprecated


ejadib:
This should be closed. We removed the LUIS Model URL in https://github.com/Microsoft/BotBuilder-Samples/commit/39d186248c044da7fff39ebe2c6fe7a26dc53194 since now it should be provided by the user.

PR #150, opened 2017-06-22
Use Top scoring Intent instead of first intent

mansoorcheema:
LUIS Result from a query returns intents and entities and scores. The code used the firstOrDefault Intent instead of the most likely intent which is TopScoringIntent. Also checking intent for null value is important to prevent crash on null exception when no intent is returned.


msft-shahins:
@ejadib can you take a look at this PR and comment?

PR #205, opened 2017-10-16
added domain - sample doesn't work without it

v-geberr:
verifying LUIS samples - this sample doesn't work without domain


stevengum97:
The LuisModel defaults to the westus endpoint, and if we're going to specify that an endpoint should be provided, we should also note that this endpoint should be specified if an app is deployed to any region _other_ than westus.

PR #207, opened 2017-10-22
Fix LUIS_MODEL.json so it can be imported successfully.

day-jeff:
LUIS app import fails because 'datetime' has been replaced with 'datetimeV2'. Fixed LUIS_MODEL.json to specify 'datetimeV2'.

### DIFF
--- a/CSharp/Blog-LUISActionBinding/LUIS_MODEL.json
+++ b/CSharp/Blog-LUISActionBinding/LUIS_MODEL.json
@@ -45,13 +45,13 @@
     {
       "name": "Checkin",
       "children": [
-        "datetime"
+        "datetimeV2"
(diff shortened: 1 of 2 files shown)


stevengum97:
Can you fix the merge conflicts? Thanks!

wenting-zhao:
Encounter this issue again, and it would be great to have a fix so that people can use the model without running into errors.  I think the conflict is just the corrected commit differs from the latest commit (``datetimeV2'' vs ``datetimev2''; the former is correct).  I could start a new pull request or someone could quickly resolve the conflict here?

Thanks!

Annabear1103:
LUIS_MODEL.json with datetimev2 is not workable now.

PR #220, opened 2017-11-21
[Node.js Sample] added listStyle to choice prompt

Wenish:
added the line so prompt choice will render as buttons in the chat.

in the node js doc its describe like buttons should be rendered
(https://github.com/Microsoft/BotBuilder-Samples/tree/master/Node/core-MultiDialogs#outcome)

c# example already does this


stevengum97:
Can you provide a link to the documentation you've mentioned?

Wenish:
@stevengum97 Its how the Message of the Bot gets rendered with Buttons.
Here the link to the Doc: 
https://github.com/Microsoft/BotBuilder-Samples/tree/master/Node/core-MultiDialogs#outcome

Wenish:
ping

PR #243, opened 2018-02-18
Update LuisBot.json

joaosamuelp:
Fix error: BadArgument: Version ID cannot be null or empty


jkolonko:
Fix for https://github.com/Microsoft/BotBuilder-Samples/issues/251

Issue #251, opened 2018-03-15
[intelligence-LUIS] Import New App error "BadArgument: Version ID cannot be null or empty"


henriquebp87:
I tried to import 'LuisBot.json' file to create a new app on my LUIS account and got the error:
"BadArgument: Version ID cannot be null or empty"
![untitled](https://user-images.githubusercontent.com/6232469/37471323-401ef5f6-2848-11e8-8448-197473052c51.png)

tiagozini:
The same with me.

gunjankhanwilkar:
Add this property "versionId": "0.1" in LuisBot.json file.
This will solve the problem and you will be able to import the file.

sgellock:
closed due to inactivity

PR #253, opened 2018-03-18
Fix LUIS model ('Version ID cannot be null or empty' when importing JSON)

jkolonko:
Fix for https://github.com/Microsoft/BotBuilder-Samples/issues/251


kiser0606:
ID

PR #261, opened 2018-04-30
Updating AzureDeploy.Json to allow RBAC Subscriptions owners to deploy to Azure App Service.

FinVamp1:
The existing template gives an error when users who are owners on the subscription try to deploy. 
Users who are in this state would see this message in the activity log.

This change updates the API versions and moves the App Settings into the main site properties.
This allows both the RBAC create to succeed and also to have the SourceControls element get the Project hint  from App Settings to build correctly in Kudu.


FinVamp1:
STATUSMESSAGE{
  "error": {
    "code": "BadRequest",
    "message": ""
  }
}

PR #301, opened 2018-08-31
QnaMaker ASP Net Core sample

daveta:
Asp.net Core using QnaMaker.  Per https://github.com/Microsoft/botbuilder-dotnet/issues/834, this is:

  *11.QnaMaker*

### Notes

**BotConfig**
- Stolen from Lamills Enterprise Bot Template

**StyleCop**
- Applied copy of our BotBuilder rules.

**AppInsights**
- Log Qna results per Darren guidance

**QnA**
- Converted to new format (.qna file). 

**Known Issues**
- Preview Emulator doesn't load .bot config with AppInsights (merge .bot files)
- QnA service metadata operations portions is currently down, so can't test CLI (using existing deployed model still works).


csharp_dotnetcore/11.QnAMaker/AspNetCore-QnA-Bot.ruleset
```diff
@@ -0,0 +1,86 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<RuleSet Name="Microsoft Managed Recommended Rules" Description="These rules focus on the most critical problems in your code, including potential security holes, application crashes, and other important logic and design errors. It is recommended to include this rule set in any custom rule set you create for your projects." ToolsVersion="10.0">
+  <Localization ResourceAssembly="Microsoft.VisualStudio.CodeAnalysis.RuleSets.Strings.dll" ResourceBaseName="Microsoft.VisualStudio.CodeAnalysis.RuleSets.Strings.Localized">
+    <Name Resource="MinimumRecommendedRules_Name" />
+    <Description Resource="MinimumRecommendedRules_Description" />
+  </Localization>
+  <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
+    <Rule Id="CA1001" Action="Warning" />
```
cleemullins: We're not using the ManagedCodeAnalysis tools, so I believe this entire XML Element (been years since I used that phrase!) can be removed.

csharp_dotnetcore/11.QnAMaker/QnAMaker/AppInsights/MyQnaConstants.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AspNetCore_QnA_Bot.AppInsights
+{
+    /// <summary>
+    /// The Application Insights property names that we're logging.
```
cleemullins: How about: [...] "to log." 

The "that we're logging" is a bit verbose.

csharp_dotnetcore/11.QnAMaker/QnAMaker/BotServices.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+
+using System;
+using System.Collections.Generic;
+using Microsoft.ApplicationInsights;
+using Microsoft.Bot.Builder.AI.QnA;
+
+namespace AspNetCore_QnA_Bot
```
cleemullins: Why the "AspNetCore" prefix? "QnA_Bot" seems sufficient.

csharp_dotnetcore/11.QnAMaker/QnAMaker/BotServices.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    /// <summary>
+    /// Class represents the bot's references to external services.
```
cleemullins: Grammar. Easiest fix: remove the word Class. "Represents the [...]"

csharp_dotnetcore/11.QnAMaker/QnAMaker/BotServices.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="BotServices"/> class.
+        /// </summary>
+        public BotServices()
+        {
+            // Given there can be multiple Qna services used in a single bot,
+            // Qna is represented as a Dictionary.  This is also modeled in the
+            // ".bot" file since the elements are named (string).
+            QnAServices = new Dictionary<string, QnAMaker>();
```
cleemullins: Remove this from the constructor. Move it to the property itself. (including the comment).

csharp_dotnetcore/11.QnAMaker/QnAMaker/BotServices.cs
```diff
@@ -0,0 +1,51 @@
...
+        public TelemetryClient TelemetryClient { get; set; }
+
+        /// <summary>
+        /// Gets or sets the (potential) set of Qna Services used.
+        /// This sample only uses a single Qna instance.
+        /// </summary>
+        /// <value>
+        /// A QnaMaker client instance created based on configuration in the .bot file.
+        /// </value>
+        public Dictionary<string, QnAMaker> QnAServices { get; set; }
```
cleemullins: public Dictionary<string, QnAMaker> QnAServices { get; set; } = new Dictionary<string, QnAMaker>();

cleemullins: In what instances would we want this to be set? I can envision *clearing* the dictionary, but replacing the dictionary itself seems like something we don't need. 

Consider making this a "get" only property.

cleemullins: I believe allowing the dictionary to be replaced opens up additional concerns around concurrency, as to even look at the dictionary requires a lock, as a different ASP.Net thread could be changing the property to a different dictionary instance. Easier to initialize during object construction and never allow it to be replaced.

csharp_dotnetcore/11.QnAMaker/QnAMaker/BotServices.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Use this to log new custom events/metrics/traces/etc into your
+        /// Application Insights service for later analysis.
+        /// </summary>
+        /// <value>
+        /// The Application Insights Telemetry client instance created based on configuration in the .bot file.
+        /// </value>
+        public TelemetryClient TelemetryClient { get; set; }
+
+        /// <summary>
+        /// Gets or sets the (potential) set of Qna Services used.
```
cleemullins: Qna should be "QnA" 

Also, if you remove the setter (see below), please update this comment.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Bot.Builder;
+using Microsoft.Bot.Schema;
+
+namespace AspNetCore_QnA_Bot
+{
+    /// <summary>
+    /// Represents a bot that can process incoming activities.
+    /// For each interaction from the user, an instance of this class is called.
```
cleemullins: I might consider adding more detail here describing the object lifecycle.
For each Activity received, a new instance of this class is created. Objects that are expensive to construct, or have a lifetime beyond the single Turn, should be carefully managed.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// </summary>
+    public class QnABot : IBot
```
cleemullins: The casing here is very inconsistent. We have:
QnABot
QnaBot (as the key value in the QnaMakerKey)
QnaMaker   <-- Suggest QnAMaker
QnaMakerKey  <-- Suggest QnAMakerKey
QnAServices
QnA_Bot (namespace name)
QnA Bot (in code comments below)

The product team seems to use "QnA" in their docs, so I believe we should do the same here.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        private readonly BotServices _services;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QnABot"/> class.
+        /// </summary>
+        /// <param name="services">Services configured from the ".bot" file.</param>
+        public QnABot(BotServices services)
+        {
+            _services = services;
```
cleemullins: Null Check + throw.

cleemullins: _services = services ?? throw new ArgumentNullException(nameof(services))

cleemullins: ... unless Null is expected, which from the code doesn't seem like it ever is.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        /// <summary>
+        /// Every Conversation turn for our QnA Bot will call this method. In this example,
+        /// the bot checks the Activty type to verify it's a message, and asks the QnA Maker
+        /// service if it recognizes an answer for the question given.
+        /// There are no dialogs used, since it's "single turn" processing, meaning there
+        /// are no set of decision points or set of properties that need to be gathered from
+        /// the user to successfully process the activity.
```
cleemullins: This comment is awesome, but a bit verbose. Please tighten it up a bit.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// </summary>
+        /// <param name="context">Turn scoped context containing all the data needed
```
cleemullins: Add a see tag here to the ITurnContext class.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// for processing this conversation turn. </param>
+        /// <param name="cancellationToken">A cancellation token that can be used by other objects
```
cleemullins: Add a see tag to CancellationToken

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// or threads to receive notice of cancellation.</param>
+        /// <returns>A task that represents the work queued to execute.</returns>
```
cleemullins: Add a see tag to Task.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
cleemullins: Add "(Optional)" to this.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+        public async Task OnTurnAsync(ITurnContext context, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (context.Activity.Type == ActivityTypes.Message && !context.Responded)
```
cleemullins: It's not clear from the summary comment above, nor is it clear from context in the code why the "!context.Responded" is part of this if statement. 

Unless I'm missing something when this code is called, if it's a message, it seems we pretty much always want to call QnA as part of this sample.

csharp_dotnetcore/11.QnAMaker/QnAMaker/QnABot.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                // Check QnAMaker model
+                var response = await _services.QnAServices[QnaMakerKey].GetAnswersAsync(context);
```
cleemullins: The dictionary here was never validated as valid. 

In this case, I would like to see a validation in the constructor for this class that validates the QnAService isn't null, and that it also has a value for the QnaMakerKey. 

The goal being, if the sample is somehow misconfigured or not properly setup, we should make sure the developer get a *VERY* clear exception message at the earliest opportunity.

csharp_dotnetcore/11.QnAMaker/README.md
```diff
@@ -0,0 +1,50 @@
+﻿This sample shows how to integrate QnA Maker in a simple bot with ASP.Net Core 2 with Application Insights.
```
cleemullins: [...] and Application Insights.

csharp_dotnetcore/11.QnAMaker/README.md
```diff
@@ -0,0 +1,50 @@
+﻿This sample shows how to integrate QnA Maker in a simple bot with ASP.Net Core 2 with Application Insights. 
+
+# Concepts introduced in this sample
+The [QnA Maker Service](https://www.qnamaker.a) enables you to build, train and publish a simple question and answer bot based on FAQ URLs, structured documents or editorial content in minutes.
```
cleemullins: URL is incorrect. www.qnamaker.ai.  (the URL here misses the "i")

csharp_dotnetcore/11.QnAMaker/README.md
```diff
@@ -0,0 +1,50 @@
...
+In this sample, we demonstrate how to use the QnA Maker service to answer questions based on a FAQ text file as input.
+
+# To try this sample
+
+- Clone the samples repository
+```bash
+git clone https://github.com/Microsoft/botbuilder-samples.git
+```
+
+### Prerequisites
```
cleemullins: Weird to have a ### following a #. Consider making this just a heading 2 rather than a heading 3.

csharp_dotnetcore/11.QnAMaker/README.md
```diff
@@ -0,0 +1,50 @@
...
+- Follow instructions [here](https://docs.microsoft.com/en-us/azure/cognitive-services/qnamaker/how-to/set-up-qnamaker-service-azure) to create a QnA Maker service.
+- Follow instructions [here](https://docs.microsoft.com/en-us/azure/cognitive-services/qnamaker/how-to/migrate-knowledge-base) to import the [smartLightFAQ.tsv](smartLightFAQ.tsv) to your newly created QnA Maker service.
+- Update [appsettings.json](appsettings.json) with your QnAMaker-Host, QnAMaker-KnowledgeBaseId and QnAMaker-EndpointKey. You can find this information under "Settings" tab for your QnA Maker Knowledge Base at [QnAMaker.ai](https://qnamaker.ai)
```
cleemullins: URL should be: "https://www.qnamaker.ai"

csharp_dotnetcore/11.QnAMaker/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Visual studio code
+- Open `BotBuilder-Samples\csharp_dotnetcore\11.QnAMaker` sample folder.
+- Bring up a terminal, navigate to BotBuilder-Samples\11.QnAMaker folder
+- type 'dotnet run'
+
+## Testing the bot using Bot Framework Emulator
+[Microsoft Bot Framework Emulator](https://github.com/microsoft/botframework-emulator) is a desktop application that allows bot developers to test and debug their bots on localhost or running remotely through a tunnel.
+
+- Install the Bot Framework emulator from [here](https://github.com/Microsoft/BotFramework-Emulator/releases)
```
cleemullins: This URL works. There is also an aka URL for this that might be better:
https://aka.ms/botframeworkemulator

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        /// <summary>
+        /// This method gets called by the runtime. Use this method to add services to the container.
+        /// </summary>
+        /// <param name="services">Specifies the contract for a collection of service descriptors.</param>
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // Load the connected services from .bot file
+            // BUGBUG: This needs to change to common bot file, once emulator understand appInsights type.
+            var botConfig = BotConfiguration.LoadAsync(@".\QnaBot.bot").Result;
```
cleemullins: If this is an Async method, this should be:
BotConfiguration.LoadAsync(@".\QnaBot.bot").GetAwaiter().GetResult();
... as doing this means the AggrigateException stuff won't crop up and confuse people. 

**Strongly Consider** using the the .Load (synchronous) method that was added today and should be in the Daily Build by Sunday.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            // Initialize Bot Connected Services Clients
+            var connectedServices = InitBotServices(botConfig);
+            services.AddSingleton(sp => connectedServices);
```
cleemullins: Should we move these 2 lines of code into an Extension method on services, so the code looks like
services.AddBotServices() or similar? Given how often this code will be cut/pasted, the clarity of the extension method is probably worth the trouble.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <param name="app">The application builder.  This provides the mechanisms to configure an application's request pipeline.</param>
+        /// <param name="env">Provides information about the web hosting environment an application is running in.</param>
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseDefaultFiles()
```
cleemullins: Is UseDefaultFiles() needed? From the [ASP.Net core 2 docs](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.defaultfilesextensions.usedefaultfiles?view=aspnetcore-2.1#Microsoft_AspNetCore_Builder_DefaultFilesExtensions_UseDefaultFiles_Microsoft_AspNetCore_Builder_IApplicationBuilder_), I don't think we do (but am honestly not 100% certain of that!)

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+                .UseBotFramework();
+        }
+
+        /// <summary>
+        /// Initialize the bot's references to external services.
+        ///
+        /// For example, Application Insights and QnaMaker services
+        /// are created here.  These external services are configured
+        /// using the BotConfigure class (based on the contents of your ".bot" file).
+        /// </summary>
```
cleemullins: Consider adding a comment here around object lifecycle, and how these are singletons and the IBot class is created new for each activity.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+                .UseStaticFiles()
+                .UseBotFramework();
+        }
+
+        /// <summary>
+        /// Initialize the bot's references to external services.
+        ///
+        /// For example, Application Insights and QnaMaker services
+        /// are created here.  These external services are configured
+        /// using the BotConfigure class (based on the contents of your ".bot" file).
```
cleemullins: Consider adding the relvent "see" tags to the comments.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// </summary>
+        /// <param name="config">Configuration object based on your ".bot" file.</param>
```
cleemullins: Add the see tag reference to the BotConfiguration type.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <returns>A object representing client objects to access external services the bot uses.</returns>
```
cleemullins: Add the See tag to the BotServices type.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            var connectedServices = new BotServices();
+
+            foreach (var service in config.Services)
+            {
+                switch (service.Type)
+                {
+                    case ServiceTypes.QnA:
+                        {
+                            // Create a QNA Maker that is initialized and suitable for passing
```
cleemullins: QnA Maker

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+                            // into the IBot-derived class (QnABot).
+                            // In this case, we're creating a custom class (wrapping the original
+                            // QnAMaker client) that logs the results of QnA Maker into Application
+                            // Insights for future anaysis.
+                            var qna = service as QnAMakerService;
```
cleemullins: If something was misconfigured, this "as" statment will result in the qna variable being null, and the next line will cryptically throw a null reference exception. Consider making this a hard cast, so the exception that results is clear and from the right line. 
var qna = (QnaMakerService)service;

cleemullins: ... or checking for null, and explicitly throwing with a clear exception. That's probably the "cleanest" thing to do.

csharp_dotnetcore/11.QnAMaker/Startup.cs
```diff
@@ -0,0 +1,164 @@
...
+
+                            var qnaMaker = new MyAppInsightsQnaMaker(qnaEndpoint, null, logUserName: false, logOriginalMessage: false);
+                            connectedServices.QnAServices.Add(qna.Name, qnaMaker);
+
+                            break;
+                        }
+
+                    case ServiceTypes.AppInsights:
+                        {
+                            var appInsights = service as AppInsightsService;
```
cleemullins: Same casting comment.

csharp_dotnetcore/11.QnAMaker/Startup.cs
daveta: Seems like a good 4.1 issue.

csharp_dotnetcore/11.QnAMaker/appsettings.json
```diff
@@ -0,0 +1,5 @@
+{
+  //  The MicrosoftAppId and MicrosoftAppPassword is created from the Azure portal http://portal.azure.com when you created your Bot.
+  "MicrosoftAppId": "",
```
yochay: Should remove these from app setting and move them to the .bot file. Thus remove from this file

PR #307, opened 2018-09-03
Spec: add stylecop instructions and default ruleset for cs samples

carlosscastro:


SPEC.md
```diff
@@ -126,7 +126,29 @@ NOTE: <ANY NOTES ABOUT THE PREREQUISITES OR ALTERNATE THINGS TO CONSIDER TO GET
...
+To benefit from [StyleCop](https://github.com/StyleCop/StyleCop) static code analysis, all samples should have the samples-specific ruleset that can be found [here](csharp_dotnetcore/samples.ruleset).
+
+When creating a new sample, follow these steps to apply the ruleset to the sample:
+
+* Copy the [samples.ruleset](csharp_dotnetcore/samples.ruleset) file to your bot directory, at the same level as the project file.
+* Rename the file to match your bot project name. For example, if your project is AspNetCore-QnA-Bot.csproj, rename the ruleset to AspNetCore-QnA-Bot.ruleset.
+* Edit your project file and add the ruleset property groups as below, entering the name of your ruleset file, for example, AspNetCore-QnA-Bot.ruleset.
+
+```
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
```
cleemullins: These lines are not actually needed. At least I don't think they are. 

These are for the old-school "CodeAnalysis" that's baked into Visual Studio, which we're not using.

csharp_dotnetcore/samples.ruleset
```diff
@@ -0,0 +1,86 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
carlosscastro: I'm not a fan of dropping this template ruleset under csharp_core, since we'd need to have another one in csharp_webapi if we'll have webapi samples.

Other options
1) Not have a template and have people copy from other samples
2) Have a template in a folder at the root level where we'll eventually have lint config for python, javascript, etc. Not sure what a good name would be for such a folder. artifacts? build? tools?

csharp_dotnetcore/samples.ruleset
```diff
@@ -0,0 +1,86 @@
...
+<RuleSet Name="Microsoft Managed Recommended Rules" Description="These rules focus on the most critical problems in your code, including potential security holes, application crashes, and other important logic and design errors. It is recommended to include this rule set in any custom rule set you create for your projects." ToolsVersion="10.0">
+  <Localization ResourceAssembly="Microsoft.VisualStudio.CodeAnalysis.RuleSets.Strings.dll" ResourceBaseName="Microsoft.VisualStudio.CodeAnalysis.RuleSets.Strings.Localized">
+    <Name Resource="MinimumRecommendedRules_Name" />
+    <Description Resource="MinimumRecommendedRules_Description" />
+  </Localization>
+  <Rules AnalyzerId="Microsoft.Analyzers.ManagedCodeAnalysis" RuleNamespace="Microsoft.Rules.Managed">
```
cleemullins: The ManagedCodeAnalysis rules (this particular XML Element) are not needed and can all be deleted. We're not using that suite of tools, and are instead using the newer stack.

csharp_dotnetcore/samples.ruleset
```diff
@@ -0,0 +1,86 @@
...
+    <Rule Id="CA2235" Action="Warning" />
+    <Rule Id="CA2236" Action="Warning" />
+    <Rule Id="CA2237" Action="Warning" />
+    <Rule Id="CA2238" Action="Warning" />
+    <Rule Id="CA2240" Action="Warning" />
+    <Rule Id="CA2241" Action="Warning" />
+    <Rule Id="CA2242" Action="Warning" />
+  </Rules>
+  <Rules AnalyzerId="StyleCop.Analyzers" RuleNamespace="StyleCop.Analyzers">
+    <Rule Id="SA1011" Action="None" />
```
cleemullins: We need these rules. :)

csharp_dotnetcore/samples.ruleset
```diff
@@ -0,0 +1,86 @@
...
+    <Rule Id="SA1129" Action="None" />
+    <Rule Id="SA1305" Action="Warning" />
+    <Rule Id="SA1309" Action="None" />
+    <Rule Id="SA1412" Action="Warning" />
+    <Rule Id="SA1600" Action="None" />
+    <Rule Id="SA1609" Action="Warning" />
+    <Rule Id="SA1633" Action="None" />
+  </Rules>
+  <Rules AnalyzerId="AsyncUsageAnalyzers" RuleNamespace="AsyncUsageAnalyzers">
+    <Rule Id="AvoidAsyncVoid" Action="Warning" />
```
cleemullins: We need these rules. :)

SPEC.md
carlosscastro: Agreed, they are not needed *as long as* the name of the file matches the project name IIRC.

csharp_dotnetcore/samples.ruleset
carlosscastro: All right, this is from Dave's PR 301 original ruleset. I'll copy the latest from here which seems to be correct and doesn't have the managed code rules: https://github.com/Microsoft/BotBuilder-Samples/pull/301/files

csharp_dotnetcore/samples.ruleset
carlosscastro: I copied the latest ruleset from PR 301 that we merged in from Dave. Sorry, I'm not sure what you mean by "we need these rules" want me to change the action or is it good as-is with the new file?

SPEC.md
carlosscastro: I just dropped it

csharp_dotnetcore/samples.ruleset
cleemullins: "Good as is." for those XmlElements. :) Was trying to be clear while my kids were climbing on me and distracting me...

carlosscastro: @cleemullins you made me lol. Thanks for the quick review!

PR #318, opened 2018-09-04
Sample: C# NLP with Dispatch

carlosscastro:
Leaving a first draft of NLP with dispatch. 

Definitely expecting some get some feedback on comments, documentation, etc. Some steps done:

- Tested documentation steps from scratch to generate luis models, qna, train, publish, dispatch, etc
- Aligned with other C# Samples (Qna, LUIS)
- Aligned in behavior with its corresponding Node sample (14.nlp-with-dispatch)
- Stylecop configured with our custom ruleset


carlosscastro:
Against master was wrong, doing another one against v4

PR #332, opened 2018-09-06
Auth sample c#

JasonSowers:
We addressed Steven's concerns by adding another sample that is simpler 👍


csharp_dotnetcore/19.Bot-Authentication/AuthenticationBot.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// <returns>A <see cref="Task"/> representing the operation result of the operation.</returns>
+        public async Task OnTurnAsync(ITurnContext turnContext, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            DialogContext dc = null;
+
+            switch (turnContext.Activity.Type)
+            {
+                case ActivityTypes.Message:
+
+                    dc = await ProcessInputAsync(turnContext, dc, cancellationToken);
```
stevengum: Why are you passing in `dc` which is a `null` value at this point? [`ProcessInputAsync()` immediately creates a DialogContext and assigns the value to `dc`](https://github.com/Microsoft/BotBuilder-Samples/pull/332/files#diff-b99c9232b9023815898ad8b91e46aa02R143), meaning that even if `dc` were not a null value, its actual value would be irrelevant.

csharp_dotnetcore/19.Bot-Authentication/AuthenticationBot.cs
```diff
@@ -0,0 +1,249 @@
...
+        /// Processes the user's input and routes the user to the appropriate step.
+        /// </summary>
+        /// <param name="turnContext">Provides the <see cref="ITurnContext"/> for the turn of the bot.</param>
+        /// <param name="dc">A <see cref="DialogContext"/> provides context for the current dialog.</param>
+        /// <param name="cancellationToken" >(Optional) A <see cref="CancellationToken"/> that can be used by other objects
+        /// or threads to receive notice of cancellation.</param>
+        /// <returns>A <see cref="Task"/> representing the operation result of the operation.</returns>
+        private async Task<DialogContext> ProcessInputAsync(ITurnContext turnContext, DialogContext dc, CancellationToken cancellationToken)
+        {
+            dc = await this._dialogs.CreateContextAsync(turnContext, cancellationToken);
```
stevengum: This line immediately stamps out whatever value dc might have. Why?

csharp_dotnetcore/19.Bot-Authentication/AuthenticationBot.cs
```diff
@@ -0,0 +1,249 @@
...
+                if (tokenResponse?.Token != null)
+                {
+                    var parts = this._stateAccessors.CommandState.GetAsync(dc.Context, cancellationToken: cancellationToken).Result.Split(' ');
+
+                    if (parts[0].ToLowerInvariant() == "me")
+                    {
+                        await OAuthHelpers.ListMeAsync(dc.Context, tokenResponse);
+                        await this._stateAccessors.CommandState.DeleteAsync(dc.Context, cancellationToken);
+                    }
+                    else if (parts[0].ToLowerInvariant().StartsWith("send"))
```
stevengum: This seems brittle.

PR #342, opened 2018-09-07
Bots for discussion and review

JonathanFingold:
**Do not merge!!!**

These are for discussion of what should be best practices.

PS. I couldn't get these to run against today's build, so they use yesterday's.


jofing-forDevReview/ComponentDialogs/Dialogs/CheckInDialog.cs
```diff
@@ -0,0 +1,92 @@
...
+            AddDialog(new WaterfallDialog(Id, new WaterfallStep[]
+            {
+                async (dc, step, cancellationToken) =>
+                {
+                    // Initialize guest information and prompt for the guest's name.
+                    step.Values[Values.GuestInfo] = new GuestInfo();
+                    return await dc.PromptAsync(PromptIds.GuestName, new PromptOptions
+                    {
+                        Prompt = MessageFactory.Text("What is your name?")
+                    });
```
tomlm: we should not use inline lamdas in waterfalls, it is clearer if these are pulled out into methods

jofing-forDevReview/UserDataBot/UserDataBot.cs
```diff
@@ -0,0 +1,110 @@
...
+    using Microsoft.Bot.Builder.Dialogs;
+    using Microsoft.Bot.Builder.Integration;
+    using Microsoft.Bot.Schema;
+    using Microsoft.Extensions.Options;
+
+    /// <summary>Defines the bot for the persisting user data tutorial.</summary>
+    public class UserDataBot : IBot
+    {
+        /// <summary>The state property accessor for user data.</summary>
+        private IStatePropertyAccessor<UserData> UserDataAccessor { get; }
```
tomlm: this is wrong

jofing-forDevReview/UserDataBot/Startup.cs
```diff
@@ -0,0 +1,93 @@
...
+                var conversationState = stateSet.BotStates.OfType<ConversationState>().FirstOrDefault();
+                return conversationState.CreateProperty<DialogState>("UserDataBot.DialogState");
+            });
+
+            // Register the main dialog set.
+            services.AddSingleton(sp =>
+            {
+                var dialogState = sp.GetRequiredService<IStatePropertyAccessor<DialogState>>();
+                return new GreetingsDialog(dialogState);
+            });
```
tomlm: this is an anti pattern, we should not do this.

jofing-forDevReview/UserDataBot/Startup.cs
```diff
@@ -0,0 +1,93 @@
...
+                IStorage storage = new CosmosDbStorage(
+                    new CosmosDbStorageOptions
+                    {
+                        AuthKey = CosmosSettings["AuthenticationKey"],
+                        CollectionId = CosmosSettings["CollectionID"],
+                        CosmosDBEndpoint = new Uri(CosmosSettings["EndpointUri"]),
+                        DatabaseId = CosmosSettings["DatabaseID"],
+                    });
+                var conversationState = new ConversationState(storage);
+                var userState = new UserState(storage);
```
tomlm: >userState [](start = 20, length = 9)

add this as singleton in DI

jofing-forDevReview/UserDataBot/Startup.cs
```diff
@@ -0,0 +1,93 @@
...
+                var CosmosSettings = Configuration.GetSection("CosmosDB");
+                IStorage storage = new CosmosDbStorage(
+                    new CosmosDbStorageOptions
+                    {
+                        AuthKey = CosmosSettings["AuthenticationKey"],
+                        CollectionId = CosmosSettings["CollectionID"],
+                        CosmosDBEndpoint = new Uri(CosmosSettings["EndpointUri"]),
+                        DatabaseId = CosmosSettings["DatabaseID"],
+                    });
+                var conversationState = new ConversationState(storage);
```
tomlm: >conversationState [](start = 20, length = 17)

add as singleton in DI

jofing-forDevReview/UserDataBot/UserDataBot.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// <summary>The dialog set that has the dialog to use.</summary>
+        private GreetingsDialog GreetingsDialog { get; }
+
+        /// <summary>Create a new instance of the bot.</summary>
+        /// <param name="options">The options to use for our app.</param>
+        /// <param name="greetingsDialog">An instance of the dialog set.</param>
+        public UserDataBot(IOptions<BotFrameworkOptions> options, GreetingsDialog greetingsDialog)
```
tomlm: >IOptions<BotFrameworkOptions> options, GreetingsDialog greetingsDialog) [](start = 27, length = 71)

this should be simply
UserDataBot(UserState, ConversationState)

Then create your property using that.

jofing-forDevReview/ComponentDialogs/Dialogs/CheckInDialog.cs
JonathanFingold: Nearly all of our current dialog samples use lambdas for waterfall steps, partly because it is less wordy and easier to read as rendered on the MS docs site. I'll bring this up in our stand up on Monday.

johnataylor: In the multi-turn bot which is a demonstration of how to construct a waterfall I pulled the logic out into separate functions. However, in the Custom-Dialog samples I just wanted a two step waterfall to drive the scenario.

In other words I think both are valid but if its a non-trivial waterfall then I think I'd agree with Tom and pull it into functions. I think, however, its actually good to show people that both techniques are valid and both can be good C#.

One interesting observation is that I find it quite hard to name the functions in a waterfall. The problem is the function itself doesn't really align with a concept in the domain. Its not exactly an event handler, nor is it a processing step as you might imagine it in a flow chart. In a step function you are processing the results of a previous step and starting anew step. We've seen people get confused between prompt validation functions and waterfall steps functions. I think the confusing is reasonable; this is an odd way to write code.

jofing-forDevReview/UserDataBot/Startup.cs
johnataylor: We should really try hard to not put business logic in Startup.cs. Its really intended for configuration at start up time. As an optimization I could imagine some common code being moved out of a Bot constructor and into Startup but as a general pattern its not right.

Its a shame the IBot instance is explicitly created in the Startup.cs code. That's not how MVC works and shouldn't be necessary in our world either.

jofing-forDevReview/ComponentDialogs/Startup.cs
```diff
@@ -0,0 +1,87 @@
...
+                var userState = new UserState(dataStore);
+                options.Middleware.Add(new BotStateSet(convState, userState));
+            });
+
+            // Create and register the dialog state accessor off of conversation state.
+            services.AddSingleton(sp =>
+            {
+                var options = sp.GetRequiredService<IOptions<BotFrameworkOptions>>().Value;
+                var stateSet = options.Middleware.OfType<BotStateSet>().FirstOrDefault();
+                var convState = stateSet.BotStates.OfType<ConversationState>().FirstOrDefault();
```
daveta: >stateSet [](start = 32, length = 8)

Not sure that i'd interrogate the BotStateSet middleware component to discover your State providers (in case people choose not to use BotStateSet).  

Instead use Options (which is new, and for this purpose):
   options.State.Add(convState);
   options.State.Add(userState);

jofing-forDevReview/ComponentDialogs/Startup.cs
```diff
@@ -0,0 +1,87 @@
...
+                return convState.CreateProperty<DialogState>("IntegratedDialogs.DialogState");
+            });
+
+            // Create and register the user profile state accessor off of user state.
+            services.AddSingleton(sp =>
+            {
+                var options = sp.GetRequiredService<IOptions<BotFrameworkOptions>>().Value;
+                var stateSet = options.Middleware.OfType<BotStateSet>().FirstOrDefault();
```
daveta: >tateSet = options.Middleware.OfType<BotStateSet>().FirstOrDefault(); [](start = 21, length = 68)

Same as above.

jofing-forDevReview/ComponentDialogs/Startup.cs
johnataylor: why do we even need to options.State at all? Why can't we just DI the state objects into the Bot object - wouldn't that just be simpler?

JonathanFingold: @daveta , I missed this API change. I'll update my code to use `options.State`.

@johnataylor , I think the state middleware still needs to be registered on the adapter, thus adding it first to the options.

PR #348, opened 2018-09-08
add bot style guidelines

tomlm:


SPEC.md
```diff
@@ -140,4 +140,127 @@ When creating a new sample, follow these steps to apply the ruleset to the sampl
...
+
+```cs
+    public class ExampleDialog
+    {
+        public ExampleDialog(BotState state)
+        {
+            this.AlarmsProperty = state.CreateProperty<Alarm[]>("Alarms");
+        }
+```
+
```
vishwacsena: @tomlm - I do not follow this one. If I use BotState for Alarms but in reality Alarms should be tied to a specific user, I'd have expected to see Alarms under userState. 

I also dont yet know the kind of information I'd want to keep in Bot State - what pieces of data relevant to my bot do you believe need to live beyond the scope of the current conversation or user? 

May be a different example that demonstrates why and what I'd want to keep in each state scopes (Bot/ User/ Conversation) would be really helpful.

PR #343, opened 2018-09-07
custom dialog implementation

johnataylor:
slot filling dialog as an example of custom dialogs and composition

PR #357, opened 2018-09-11
Add custom dialog sample for NodeJS

benbrown:
This is a port of @johnataylor's #343 in Node.

TODO:
 - [x] update inline comments
- [x] update readme


benbrown:
It probably makes the most sense to have @johnataylor review this FIRST since this is a direct port of his C# version.

benbrown:
This got rolled into #363

PR #360, opened 2018-09-11
Bug fixes

JonathanFingold:
Fixes:

- Pipe issue with conversation IDs.
- Failing to write when completing a "job".


JonathanFingold:
@daveta , I'm not convinced my "quick and dirty" fix for the pipe issue is the correct one.

PR #365, opened 2018-09-12
BasicBot Asp Net Core 2

daveta:
Basic bot template that puts together cards, NLP (LUIS).
- https://github.com/Microsoft/botbuilder-dotnet/issues/834
- 13.Basic-Bot-Template

@johnataylor : If you can take a look at GreetingDialog.cs / DisplayGreetingStateStepAsync.
I am not sure what the prevailing winds are with manual saves.  I'm sure there's a better spot/way to do this.
@sgellock : Made some slight mods from the original.  After the bot has the name/city, it's kind of lonely so added "$"Hi `{stepContext.Values[NameValue]}`, living in `{stepContext.Values[CityValue]}`,"
                    + " I understand greetings and asking for help!  Or start your connection over for a card." to give some guidance.


csharp_dotnetcore/13.Basic-Bot-Template/BasicBotAccessors.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Microsoft.BotBuilderSamples
+{
+    /// <summary>
+    /// This class is created as a Singleton and passed into the IBot-derived constructor.
+    ///  - See <see cref="BasicBot"/> constructor for how that is injected.
+    ///  - See the Startup.cs file for more details on creating the Singleton that gets
+    ///    injected into the constructor.
+    /// </summary>
+    public class BasicBotAccessors
```
tomlm: >BasicBotAccessors [](start = 17, length = 17)

I really really really don't like this as a pattern.  It feels like an anti pattern to have a global singleton with all of your accesors on it. 

What is the case for these things not being defined as close to the code that consumes it as possible?  

* DialogState should be defined in the bot where the dialogSet is defined
* GreetingState should be definedin the dialog which does greeting 
etc.

csharp_dotnetcore/13.Basic-Bot-Template/Startup.cs
```diff
@@ -0,0 +1,217 @@
...
+                // Create the custom state accessor.
+                // State accessors enable other components to read and write individual properties of state.
+                var accessors = new BasicBotAccessors(userState, conversationState)
+                {
+                    GreetingStateProperty = userState.CreateProperty<GreetingState>(BasicBotAccessors.GreetingStateName),
+                    DialogStateProperty = conversationState.CreateProperty<DialogState>(BasicBotAccessors.DialogStateName),
+                };
+
+                return accessors;
+            });
```
tomlm: This feels like an anti pattern.  What is the case for having to push your bot logic into startup.cs?

csharp_dotnetcore/13.Basic-Bot-Template/Dialogs/MainDialog/MainDialog.cs
```diff
@@ -0,0 +1,203 @@
...
+                        var luisResults = await _services.LuisServices[LuisKey].RecognizeAsync(dc.Context, cancellationToken).ConfigureAwait(false);
+                        var topIntent = luisResults?.GetTopScoringIntent();
+                        var interruptResult = InterruptionStatus.NoAction;
+                        if (topIntent != null)
+                        {
+                            // See if there are any conversation interrupts we need to handle
+                            switch (topIntent.Value.intent)
+                            {
+                                case GreetingIntent:
+                                    if (topIntent.Value.score > .5)
```
johnataylor: We should not show code that compares scores to constants. Actually we should only compare scores within a model.

In this particular case we are hitting up against the issue where LUIS is not going to work well for discriminating single word utterances. So for discriminating intent between, say, "cancel," "help" and "hi" we should not use LUIS instead we could use something like regular expressions or some other pattern match.

csharp_dotnetcore/13.Basic-Bot-Template/Startup.cs
```diff
@@ -0,0 +1,217 @@
...
+                            luisServices.Add(luis.Name, recognizer);
+                            break;
+                        }
+                }
+            }
+
+            var connectedServices = new BotServices(luisServices);
+            return connectedServices;
+        }
+    }
```
tomlm: why isn't this in the bot?  This feels like bot stuff

csharp_dotnetcore/13.Basic-Bot-Template/BasicBotAccessors.cs
johnataylor: I'm not sure I like this pattern that much either. But at least we are consistent so if we want to change it we should be able to do that mechanically across all the samples.

csharp_dotnetcore/13.Basic-Bot-Template/Dialogs/Greeting/GreetingDialog.cs
```diff
@@ -0,0 +1,169 @@
...
+                var options = new PromptOptions
+                {
+                    Prompt = new Activity
+                    {
+                        Type = ActivityTypes.Message,
+                        Text = "What is your name ?",
+                    },
+                };
+
+                return await dc.PromptAsync(nameof(NamePrompt), options).ConfigureAwait(false);
```
tomlm: can't this simply be:
await dc.PromptAsync(nameof(NamePrompt), "what is your name?", options) ?

Or am I missing something?

daveta:
Going to re-open this PR - just needed a snapshot for AndreaO.

csharp_dotnetcore/13.Basic-Bot-Template/Dialogs/Greeting/GreetingDialog.cs
daveta: You must have overheard us, that's checked in!

csharp_dotnetcore/13.Basic-Bot-Template/Startup.cs
daveta: We want to cache the client and not re-create the connection for every turn..

csharp_dotnetcore/13.Basic-Bot-Template/Startup.cs
daveta: I could go either way on this - regardless, we're going to have to pass state.  If we want to recreate accessors on every turn, don't really care - not expensive.

csharp_dotnetcore/13.Basic-Bot-Template/BasicBotAccessors.cs
daveta: Either way.  
We still have to inject the underlying ConversationState/UserState  - if we want to remove accessors, it's not a big runtime hit to recreate accessors on every turn.

PR #366, opened 2018-09-12
Modify to read from .bot config for appId/appPassword for vishwac

daveta:
@vishwacsena and @Andrea-Orimoto wanted a sample Echobot to read from .bot config file (instead of using appsettings.json).

### DIFF
--- a/csharp_dotnetcore/2.EchoBot-With-Counter/Startup.cs
+++ b/csharp_dotnetcore/2.EchoBot-With-Counter/Startup.cs
@@ -9,6 +9,8 @@ using Microsoft.Bot.Builder;
 using Microsoft.Bot.Builder.BotFramework;
 using Microsoft.Bot.Builder.Integration;
 using Microsoft.Bot.Builder.Integration.AspNet.Core;
+using Microsoft.Bot.Configuration;
+using Microsoft.Bot.Connector.Authentication;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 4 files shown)


csharp_dotnetcore/2.EchoBot-With-Counter/Startup.cs
```diff
@@ -48,7 +50,10 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddBot<EchoWithCounterBot>(options =>
             {
-                options.CredentialProvider = new ConfigurationCredentialProvider(Configuration);
+                // Load the connected services from .bot file.
+                var botConfig = BotConfiguration.Load(@".\EchoBotWithCounter.bot");
+                var endpointService = (EndpointService)botConfig.Services.First(s => s.Type == "endpoint");
```
cleemullins: There is quite a bit that can go wrong here. For samples, I would put checks + explicit exceptions around this.

cleemullins: .FirstOrDefault(). First will throw if no endpoints exists, which probably isn't what's desired here.

cleemullins: Why is a cast to EndpointService needed? Can we update the code so this isn't needed?

cleemullins: ... especially as I see your have ? operators on the endpoint later, indicating you're expecting and good with null values coming back.

PR #370, opened 2018-09-13
Bot Configuration Consistency across samples using .bot

carlosscastro:
Consistent configuration, app id and pwd from .bot file using SimpleconfigurationProvider.

Fixes: https://github.com/Microsoft/botbuilder-dotnet/issues/912, except for enterprise bot.


carlosscastro:
@cleemullins Good idea! I just added a change that includes your suggestions. All call the method, but QnA and a few other special cases also load the config for other things. This could be refactored further, maybe add a LoadBotConfig method so that loading is unified. I'll try to get to it but it's looking much better already.

csharp_dotnetcore/10.Prompt-Validations/Startup.cs
```diff
@@ -91,5 +93,14 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
...
         }
+
+        /// <summary>
+        /// Initializes the credential provider, using by default the <see cref="SimpleCredentialProvider"/>.
+        /// </summary>
+        /// <param name="options"><see cref="BotFrameworkOptions"/> for the current bot.</param>
+        private static void InitCredentialProvider(BotFrameworkOptions options)
+        {
+            InitCredentialProvider(options);
+        }
```
cleemullins: This will go kaboom. :)

csharp_dotnetcore/2.EchoBot-With-Counter/EchoBotWithCounter.csproj
```diff
@@ -7,7 +7,7 @@
 
 
   <ItemGroup>
-    <Content Include="EchoBotWithCounter.bot">
+    <Content Include="BotConfiguration.bot">
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
cleemullins: This sample doesn't have any modifications to startup.cs, as all the other samples do. 

Did you miss this one, or is this an oversight?

csharp_dotnetcore/10.Prompt-Validations/Startup.cs
carlosscastro: LOL

csharp_dotnetcore/2.EchoBot-With-Counter/EchoBotWithCounter.csproj
carlosscastro: Good catch!!!
Oversight

Issue #373, opened 2018-09-13
v4 C# samples should explicitly call the state accessor's SetAsync method


JonathanFingold:
### SDK Platform
C# (ASP.NET Core)

### URL of Code Sample
csharp_dotnetcore/2.EchoBot-With-Counter/

However, many (if not most) of the samples that work with state seem to skip the SetAsync step.

### Problem Description
Update sample 2 to use an `int` for the state property, instead of the CounterState class.
Run the sample.
Note that in the Emulator the counter always stays at 1.

By not calling SetAsync, we're relying on not losing the original reference to our state object. This can get broken in many ways, but using a value type as the state object is the easiest.

JonathanFingold:
@johnataylor , @cleemullins , Can one of you surface this. It seems like a key concept for devs to understand.

daveta:
@JonathanFingold : Yep, it's not setting.

daveta:
Thanks Jonathan!

PR #379, opened 2018-09-15
Echobot Fixes

daveta:
- Asp.net Core attempting to load wrong bot config!
 Apparently we are standardizing on "BotConfiguration.bot" - but code was pointing at old name.  We probably need to check other samples.
- Fix BotConfiguration.bot endpoint name
We need to name local endpoint in bot config "development" to work with Azure deployment stuff.  This probably should be adopted across the board.
- Update packages on WebApi
 Latest Package references
 Remove remnants of Application Insights
- Fix set on accessor
 Set wasn't called on accessor!


cleemullins:
@daveta Thanks Dave. Not sure how @carlosscastro & I both missed the botconfig change in this sample! 

This also clarifies #373, and provides a good pattern to follow.

PR #389, opened 2018-09-16
Fixing a bug with the translator middleware

ovishesh:
Fixing a bug where the request to change from Spanish to English was not being met because "es" was being translated to "In".


cleemullins:
@carlosscastro Can you take a quick look? Looks like a fix we should take ASAP. This same problem *probably* exist in the C# version of the sample and would need to be fixed there as well.

stevengum:
@ovishesh thanks for reporting this, I took this change in PR #512 while finishing the Node.js sample.  @carlosscastro once you take a look at this, can you indicate if you want another fix in place of Vishesh's contribution?

PR #411, opened 2018-09-18
fixes

JasonSowers:
I happened to need to do some testing on these samples and they were not updated to incorporate the changes to the SDK that were made when I was out.

### DIFF
--- a/csharp_dotnetcore/84.Bot-Authentication-MSGraph/Startup.cs
+++ b/csharp_dotnetcore/84.Bot-Authentication-MSGraph/Startup.cs
@@ -64,11 +64,8 @@ namespace Microsoft.BotBuilderSamples
                 options.State.Add(convoState);
                 options.State.Add(userState);
 
-                // Add State to BotStateSet Middleware (that require auto-save)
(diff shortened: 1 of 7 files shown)


csharp_dotnetcore/84.Bot-Authentication-MSGraph/Startup.cs
```diff
@@ -64,11 +64,8 @@ public void ConfigureServices(IServiceCollection services)
                 options.State.Add(convoState);
                 options.State.Add(userState);
 
-                // Add State to BotStateSet Middleware (that require auto-save)
-                // The BotStateSet Middleware forces state storage to auto-save when the Bot is complete processing the message.
-                // Note: Developers may choose not to add all the State providers to this Middleware if save is not required.
-                var stateSet = new BotStateSet(options.State.ToArray());
-                options.Middleware.Add(stateSet);
+                // Automatically save state at the end of the turn
+                options.Middleware.Add(new AutoSaveStateMiddleware(convoState, userState));
```
cleemullins: We should remove this and explicity save state in the turn.

PR #429, opened 2018-09-20
Finalize javascript_nodejs/10.prompt-validations

benbrown:
- [x] Include all DCR changes
- [x] Explicit state save
- [x] onTurnError at adapter
- [x] README to include
   - [x] msbot clone steps
- [x] ~C# sample -- Include .bot file in the build output~
- [x] remove and sort usings
- [x] ~remove azuredeploy.* under deploymentscripts folder (if any)~


benbrown:
Applied all previous tune-ups to this one!

samples/javascript_nodejs/10.prompt-validations/.env
```diff
@@ -0,0 +1 @@
+botFilePath=prompt-validations-bot.bot
```
vishwacsena: Let's add botFileSecret=

samples/javascript_nodejs/10.prompt-validations/bot.js
```diff
@@ -64,7 +64,7 @@ class MainDialog {
             // If no response has been sent, start the onboarding dialog.
             if (!turnContext.responded) {
                 await dc.beginDialog(ONBOARD_USER);
-            } 
+            }
         } else if (
             turnContext.activity.type === ActivityTypes.ConversationUpdate &&
```
vishwacsena: This needs to be cleaned up. Welcoming users section is not correct. Can you bring this section over from the basic-bot sample?

samples/javascript_nodejs/10.prompt-validations/prompts/dobPrompt/index.js
```diff
@@ -7,7 +7,7 @@ const DATE_LOW_BOUNDS = new Date('1918-08-24');
 const DATE_HIGH_BOUNDS = new Date('2018-08-24');
 
 // This is a custom DateTimePrompt that requires the date to be between DATE_LOW_BOUNDS and DATE_HIGH_BOUNDS.
-module.exports = class DOBPrompt extends DateTimePrompt {
+module.exports.DobPrompt = class DobPrompt extends DateTimePrompt {
     constructor(dialogId) {
         super(dialogId, async (prompt) => {
             try {
```
vishwacsena: We should add a comment here to point readers to our text-recognizers suite .. 
https://github.com/Microsoft/Recognizers-Text/tree/master/JavaScript/packages/recognizers-text-suite

may be post ignite we can update this to show using the Timex and resolvers with constraints to have more impactful validation rules.. 

See [here](https://github.com/Microsoft/BotBuilder-Samples/blob/vishwac/cafebotv2/javascript_nodejs/50.contoso-cafe-bot/dialogs/shared/stateProperties/reservationProperty.js) for an usecase in cafebot.

benbrown: good idea, it is an impressive library

samples/javascript_nodejs/10.prompt-validations/bot.js
benbrown: Not sure what you want me to do to clean it up.  

If we bring the other class in, that's fine, but would effectively replace this sample which demonstrates multiple different type of custom prompt validators, whereas that one only has string validators.

But I'd like to know what you don't like about how I did this too!

benbrown: Updated!

PR #434, opened 2018-09-20
dotnetcore/02.echo-with-counter

JasonSowers:
--
- [x] Include all DCR changes
- [ ] Explicit state save
- [ ] Method name changes begin, end etc
- [ ] Does not use mainDialog unless that file actually extends Dialog/ ComponentDialog
- [ ] Compiles with latest packages
- [ ] Runs per instructions in README
- [ ] Add commented out code to use blob storage
- [ ] Update to look for production/ local (bot configuration)
- [ ] onTurnError at adapter
- [ ] Include .env or appsettings.json or web.config with botFilePath and botFileSecret
- [ ] Deployment scripts folder has been cleaned up and only includes msbot export files
- [ ] README to include
- [ ] msbot clone steps
- [ ] Tools install/ uninstall
- [ ] Concise set of steps to run the sample
- [ ] C# sample -- Include .bot file in the build output
- [ ] remove and sort usings
- [ ] remove azuredeploy.* under deploymentscripts folder (if any)
- [ ] Welcoming users is consistent & correct. See 13.basic bot for nodejs & 3. welcome user for C#


JasonSowers:
Daves PR #348 handles this

PR #441, opened 2018-09-20
Finalize Node.js v4 06.using-cards

stevengum:
- [x] Include all DCR changes
    - [x] Explicit state save
- [X] Add commented out code to use blob storage
- [X] onTurnError at adapter
- [X] README to include
    - [x] msbot clone steps
- [x] remove and sort usings


samples/javascript_nodejs/06.using-cards/index.js
```diff
@@ -0,0 +1,91 @@
...
+    adapter.processActivity(req, res, async (turnContext) => {
+        // Route to bot.
+        await bot.onTurn(turnContext);
+    });
+});
+
+// Catch-all for errors.
+adapter.onTurnError = async (context, error) => {
+    console.error(`\n [onTurnError]: ${ error }`);
+    context.sendActivity(`Oops. Something went wrong!`);
```
vishwacsena: You should clean up conversation state here.

samples/javascript_nodejs/06.using-cards/bot.js
```diff
@@ -87,10 +85,10 @@ class RichCardsBot {
 
     /**
      * Send a Rich Card response to the user based on their choice.
-     * 
+     *
      * This method is only called when a valid prompt response is parsed from the user's response to the ChoicePrompt.
-     * @param {TurnContext} turnContext
-     * @param {DialogTurnResult} dialogTurnResult 
+     * @param {Object} turnContext
```
vishwacsena: Let's go back and not have generic objects here. This helps with readability even if not with intellisense.

PR #446, opened 2018-09-20
c#/04.simple-prompt

JasonSowers:
--
- [x] Include all DCR changes
- [x] Explicit state save
- [x] onTurnError at adapter
- [x] README to include
- [x] msbot clone steps
- [x] C# sample -- Include .bot file in the build output
- [x] All samples include the latest botfile (no "secretKey" field; replaced with 
 "padLock"). Use latest MSBOT off the tools MyGet feed to generate the bot file
- [x] remove and sort usings


samples/csharp_dotnetcore/04.simple-prompt/Startup.cs
```diff
@@ -0,0 +1,142 @@
...
+                    ConversationDialogState = conversationState.CreateProperty<DialogState>("DialogState"),
+                };
+
+                return accessors;
+            });
+        }
+
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
+        {
+            _isProduction = env.IsProduction();
```
vishwacsena: @JasonSowers  - talk to dave. This initialization needs to change in all our samples. We are consuming _isProduction before it is set. But i wouldnt block this PR due to that.

PR #448, opened 2018-09-20
Finalize javascript_nodejs/08.suggested-actions

stevengum:
- [x] Include all DCR changes
    - [x] Explicit state save
- [X] ~Add commented out code to use blob storage~
- [X] onTurnError at adapter
- [X] README to include
    - [x] msbot clone steps
- [x] remove and sort usings


samples/javascript_nodejs/08.suggested-actions/bot.js
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            await turnContext.sendActivity(`[${ turnContext.activity.type } event detected.]`);
+        }
+    }
+
+    /**
+     * Send a welcome message along with suggested actions for the user to click.
+     * @param {TurnContext} turnContext A TurnContext instance containing all the data needed for processing this conversation turn.
+     */
+    async sendWelcomeMessage(turnContext) {
```
vishwacsena: Can we simplify this? 

if (turnContext.activity.membersAdded) {
            // Iterate over all new members added to the conversation
            for (var idx in turnContext.activity.membersAdded) {
                if (turnContext.activity.membersAdded[idx].id !== turnContext.activity.recipient.id) {
                    await turnContext.sendActivity(`foo`);
                }
            }
        }

stevengum: Done!

PR #449, opened 2018-09-20
c#/05.multi-turn-prompts

JasonSowers:
--
- [x] Include all DCR changes
- [x] Explicit state save
- [x] onTurnError at adapter
- [x] README to include
- [x] msbot clone steps
- [x] C# sample -- Include .bot file in the build output
- [x] remove and sort usings


samples/csharp_dotnetcore/05.multi-turn-prompt/Startup.cs
```diff
@@ -0,0 +1,161 @@
...
+                options.CredentialProvider = new SimpleCredentialProvider(endpointService.AppId, endpointService.AppPassword);
+
+                // Creates a logger for the application to use.
+                ILogger logger = _loggerFactory.CreateLogger<MultiTurnPromptsBot>();
+
+                // Catches any errors that occur during a conversation turn and logs them.
+                options.OnTurnError = async (context, exception) =>
+                {
+                    logger.LogError($"Exception caught : {exception}");
+                    await context.SendActivityAsync("Sorry, it looks like something went wrong.");
```
vishwacsena: We should do another pass over all C# samples and add clear state and save state in onTurnError

vishwacsena:
@JasonSowers - Just fix up the default.htm and this should be good to go. Thanks!

PR #471, opened 2018-09-22
update recipe to include abs, prod endpoint, app insights, and blob

yochay:
This needs to go to generators VSIX and Yeoman


sgellock:
we've done a sweep on samples and have integrated this with each appropriate sample

PR #487, opened 2018-09-22
Dispatch nodejs 

vishwacsena:
- Dispatch sample
- Cleaned up all recipe files for samples that are already checked in.


samples/javascript_nodejs/14.nlp-with-dispatch/homeAutomation.js
```diff
@@ -47,61 +47,61 @@ class homeAutomation {
         });
     }
     /**
-     * 
-     * @param {Object} context context object
+     *
+     * @param {TurnContext} context context object
      */
-    async onTurn(context) {
+    async onTurn(turnContext) {
```
sgellock: do one of two things, either change thus parameter from turnContext to context, or change line 51 change context to turnContext in the param comment

sgellock: in other places you're using 'context' as the param name

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
```diff
@@ -0,0 +1,116 @@
...
+
+// Create adapter.
+// See https://aka.ms/about-bot-adapter to learn more about .bot file its use and bot configuration .
+const adapter = new BotFrameworkAdapter({
+    appId: endpointConfig.appId || process.env.microsoftAppID,
+    appPassword: endpointConfig.appPassword || process.env.microsoftAppPassword
+});
+
+// Catch-all for errors.
+adapter.onTurnError = async (context, error) => {
```
sgellock: do we need to work around the 4.0.6 bug and call load on the state before issuing the clear?

samples/javascript_nodejs/14.nlp-with-dispatch/home-automation-state.js
```diff
@@ -4,74 +4,76 @@ const HOME_ATUOMATION_STATE_PROPERTY = 'homeAutomation.state';
...
     }
 };
 
 class HomeAutomationState {
     /**
-     * 
-     * @param {object} convoState Conversation state
-     * @param {object} userState User state 
+     *
+     * @param {ConversationState} convoState Conversation state
```
sgellock: in the other samples we use 'conversationState' as the variable name, not 'convoState'.  That would require a change here and within this function

samples/javascript_nodejs/14.nlp-with-dispatch/homeAutomation.js
vishwacsena: Updated all to be turnContext.

vishwacsena:
@sgellock  - addressed the PR feedback. verified onTurnError works correctly.

PR #504, opened 2018-09-22
Node.js message routing update

carlosscastro:
--
- [ ] All files have been spell checked
- [ ] All style cop/ linting issues have been addressed
- [x] Include all DCR changes
- [x] Explicit state save
- [ ] onTurnError at adapter
- [ ] README to include
 
- [x] msbot clone steps
- [ ] Tools install/ uninstall
- [ ] C# sample -- Include .bot file in the build output
- [ ] remove and sort usings


carlosscastro:
@vishwacsena  I was planning on putting moving it in another PR, not sure if I should move to samples in this pr.

javascript_nodejs/09.message-routing/index.js
```diff
@@ -79,10 +79,6 @@ const memoryStorage = new MemoryStorage();
 const conversationState = new ConversationState(memoryStorage);
 const userState = new UserState(memoryStorage);
 
-// Use the BotStateSet middleware to automatically read and write conversation and user state.
```
vishwacsena: You might want to incldue onTurnError handler. See the nlp-with-dispatch bot for the correct code.

sgellock: I'd also move it into the samples folder.  I'll merge it once those two things are addressed

carlosscastro: Thanks for reviewing! Updating the PR in a few minutes :)

PR #510, opened 2018-09-23
Darrenj/remoteenttemplate

darrenj:
Removing ent template that is in the AI repo


vishwacsena:
You need to issue this to V4 branch and not master.

PR #514, opened 2018-09-23
Finalize Node.js v4 11.qnamaker sample

stevengum:
- [x] Include all DCR changes
    - [x] ~Explicit state save~
- [X] ~Add commented out code to use blob storage~
- [X] onTurnError at adapter
- [X] README to include
    - [x] msbot clone steps
- [x] remove and sort usings
- [x] uses named exports


samples/javascript_nodejs/11.qnamaker/index.js
```diff
@@ -48,20 +47,41 @@ const qnaEndpointSettings = {
...
 // Create adapter. See https://aka.ms/about-bot-adapter to learn more about adapters.
 const adapter = new BotFrameworkAdapter({
     appId: endpointConfig.appId || process.env.MicrosoftAppId,
     appPassword: endpointConfig.appPassword || process.env.MicrosoftAppPassword
 });
 
-// Create the bot that will handle incoming messages.
-const qnaMakerBot = new QnAMakerBot(qnaEndpointSettings, {});
+// Catch-all for errors.
+adapter.onTurnError = async (turnContext, error) => {
```
sgellock: onTurnError needs to use the load, clear, save pattern.  


    // Clear out state
    await conversationState.load(context);
    await conversationState.clear(context);
    // Save state changes.

stevengum: Sample does not use `ConversationState` which is why it's missing the state clearing mechanisms.

stevengum:
@sgellock I've updated the bot.recipe file and left a comment about `adapter.onTurnError()` [here](https://github.com/Microsoft/BotBuilder-Samples/pull/514#discussion_r219711640).

PR #517, opened 2018-09-23
Folder rename for #20 and #21

JasonSowers:
Renamed folders:
- 22.qna-with-appinsights =>20.qna-with-appinsights
- 23.luis-with-appinsights =>21.qna-with-appinsights


carlosscastro:
Thanks @JasonSowers !

PR #521, opened 2018-09-23
cleanup c#/06.using-cards

JasonSowers:
- move isProduction flag to startup constructor
- added membersadded check to welcome message
- added deploy to Azure section to readme


carlosscastro:
:shipit: Thanks @JasonSowers

PR #530, opened 2018-09-24
Finalize TS 12.nlp-with-luis

stevengum:
- [x] Include all DCR changes
    - [x] ~Explicit state save~
- [X] Add commented out code to use blob storage
- [X] onTurnError at adapter
- [X] README to include
    - [x] msbot clone steps
- [x] remove and sort usings
- [x] uses named exports
- [x] use updated bot.recipe


samples/javascript_typescript/12.nlp-with-luis/src/index.ts
```diff
@@ -48,7 +42,7 @@ const luisConfig = <ILuisService>botConfig.findServiceByNameOrId(LUIS_CONFIGURAT
 // Map the contents to the required format for `LuisRecognizer`.
 const luisApplication: LuisApplication = {
     applicationId: luisConfig.appId,
-    azureRegion: luisConfig.region,
+    endpoint: `https://${ luisConfig.region }.api.cognitive.microsoft.com`,
```
vishwacsena: This should instead be luisConfig.getEndpoint(). I'll fix this when I merge into master.

PR #537, opened 2018-09-24
fix incorrect isSubtle field

nyxsys:
Fixes issue where welcome card would throw an internal server error.

## Proposed Changes

Switched "isSubtle" from presumably truthy "yes" value to just `true` as msteams would throw an error 
`[onTurnError]: Error: Payload is incorrect, field is in the wrong format: isSubtle` when this field was set as "yes"

Picture of fixed card in MSTeams:
welcome card

  - 
  -
  -

## Testing
Create a bot with MSteams channel, initialize welcome bot, say "help" and welcome card should appear without internal server errors.


JasonSowers:
Tested and confirmed this bug and fix.

PR #591, opened 2018-10-01
Connection setting variable was undefined when signing out

rob-derosa:
Fixes #

## Proposed Changes

Updated connection string var to be correct when signing out

screen shot 2018-10-01 at 10 38 32 am


cleemullins:
@stevengum @JasonSowers 

Could one of you take a quick look? I don't know enough about the Authentication samples to sign-off.

PR #687, opened 2018-10-12
removed rendundant field initializer + readonly + improved LINQ query

lucaleone:
## Proposed Changes
_isProduction has been made readonly to avoid user changes.
The linq query at line 68 has been mad shorter to improver readability.


cleemullins:
@lucaleone Thanks! We probably won't push a new VSIX until the Oct 31 time frame, but thanks for the update.

PR #714, opened 2018-10-17
Ckk/fixsecret

ckkashyap:
This PR was created after running step 3 from https://help.github.com/articles/removing-sensitive-data-from-a-repository/

git filter-branch --force --index-filter "git rm --cached --ignore-unmatch csharp_dotnetcore/60.ConversationHistory/Startup.cs" --prune-empty --tag-name-filter cat -- --all

This is to remove the connectionstring that was accidentally pushed into the private branch.


vishwacsena:
2808 files changed does not look right. @ckkashyap - can you please take a look and update the PR.

PR #721, opened 2018-10-18
vstkanb message routing

### LINKED ISSUE #629: Node.js issue with MessageRoutingBot
In the Greetings extension of ComponentDialog in the greeting.js file
when intializeStateStep() is called the WaterfallStepContext is not being defined. as a result the bot fails when it tries to call options. 

```
async initializeStateStep(dc, step) {
        let greetingState = await this.greetingStateAccessor.get(dc.context);
        if (greetingState === undefined) {
            if (step.options && step.options.greetingState) {
                await this.greetingStateAccessor.set(dc.context, step.options.greetingState);
            } else {
                await this.greetingStateAccessor.set(dc.context, new GreetingState());
            }
            await this.userState.saveChanges(dc.context);
        }
        return await step.next();
    }
```

I went ahead and did a console.log of dc (step returned undefined) and got the following:

```
WaterfallStepContext {
  dialogs:
   DialogSet {
     dialogState: null,
     dialogs:
      { profileDialog: [Object],
        namePrompt: [Object],
        cityPrompt: [Object] } },
  context:
   TurnContext {
     _respondedRef: { responded: false },
     _turnState: Map { Symbol(state) => [Object] },
     _onSendActivities: [],
     _onUpdateActivity: [],
     _onDeleteActivity: [],
     _adapter:
      BotFrameworkAdapter {
        middleware: [Object],
        settings: [Object],
        credentials: [Object],
        credentialsProvider: [Object],
        isEmulatingOAuthCards: false,
        turnError: [AsyncFunction] },
     _activity:
      { type: 'message',
        text: 'hello',
        from: [Object],
        locale: 'en-US',
        textFormat: 'plain',
        timestamp: '2018-10-05T20:51:33.185Z',
        channelData: [Object],
        channelId: 'emulator',
        conversation: [Object],
        id: '712db310-c8e0-11e8-9b51-f3b28890b680',
        localTimestamp: '2018-10-05T15:51:33-05:00',
        recipient: [Object],
        serviceUrl: 'http://localhost:52543' } },
  stack: [ { id: 'profileDialog', state: [Object] } ],
  _info:
   { index: 0,
     options: {},
     reason: 'beginCalled',
     result: undefined,
     values: {},
     onNext: [Function: onNext] } }
```

The greetingStateAccessor.get() is also not returning a value back. It's a moot point however because when step.next() is called the program will fail. I checked that the waterfall was being created before it's called. I'm just not sure why it doesn't get a step context

stevkan:
Fixes #629  

## Proposed Changes
- bot.js, line 97: Corrected typo in DialogTurnStatus name.
- bot.js, line 75: Adjusted code to return 'interrupted' result into 'dialogResult' for checking against.
- greeting.js: Adjusted waterfall dialog step functions and validator functions to accept and use only one parameter. Corrected the params used.
- greeting.js, line 178: Changed dc.end() to dc.endDialog().

### DIFF
--- a/samples/javascript_nodejs/09.message-routing/bot.js
+++ b/samples/javascript_nodejs/09.message-routing/bot.js
@@ -48,7 +48,7 @@ class MessageRoutingBot {
         this.dialogs = new DialogSet(this.dialogState);
         this.dialogs.add(new GreetingDialog(GREETING_DIALOG, this.greetingStateAccessor, userState));
 
-        
+
         this.conversationState = conversationState;
         this.userState = userState;
     }
@@ -66,7 +66,7 @@ class MessageRoutingBot {
         const dc = await this.dialogs.createContext(context);
 
         if(context.activity.type === ActivityTypes.Message) {
-            
+
             const utterance = context.activity.text.trim().toLowerCase();
 
             // handle conversation interrupts first
@@ -78,7 +78,7 @@ class MessageRoutingBot {
             // Continue the current dialog
             const dialogResult = await dc.continueDialog();
 
-            // If no one has responded, 
+            // If no one has responded,
             if (!dc.context.responded) {
                 // Examine results from active dialog
                 switch (dialogResult.status) {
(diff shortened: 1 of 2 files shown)


samples/javascript_nodejs/09.message-routing/dialogs/greeting/greeting.js
```diff
@@ -125,61 +122,60 @@ class Greeting extends ComponentDialog {
...
     /**
      * Validator function to verify that user name meets required constraints.
-     * 
-     * @param {DialogContext} context for this dialog
+     *
      * @param {PromptValidatorContext} prompt context for this prompt
      */
-    async validateName(context, prompt) {
+    async validateName(prompt) {
         // Validate that the user entered a minimum lenght for their name
```
stevengum: >lenght [](start = 52, length = 6)

Can you fix these typos? There is another `lenght` typo down below in the next validator.

samples/javascript_nodejs/09.message-routing/dialogs/greeting/greeting.js
```diff
@@ -125,61 +122,60 @@ class Greeting extends ComponentDialog {
...
+        prompt.activeDialog = await this.greetingStateAccessor;
+        prompt.context = await prompt.context;
         const value = (prompt.recognized.value || '').trim();
         if (value.length >= NAME_LENGTH_MIN) {
-            prompt.end(value);
+            return true;
```
stevengum: Please add a comment here specifying why you're returning `true`. While the code in the `else` block is correct, please set an explicit `return false;` and explain why that needs to occur as well.

samples/javascript_nodejs/09.message-routing/dialogs/greeting/greeting.js
```diff
@@ -125,61 +122,60 @@ class Greeting extends ComponentDialog {
...
-     * 
+     *
      * @param {DialogContext} dc context for this dialog
      */
     async greetUser(dc) {
         const greetingState = await this.greetingStateAccessor.get(dc.context);
         // Display to the user their profile information and end dialog
         await dc.context.sendActivity(`Hi ${greetingState.name}, from ${greetingState.city}, nice to meet you!`);
-        return await dc.end();
+        return await dc.endDialog();
```
stevengum: This function takes a `WaterfallStepContext`, not a `DialogContext` instance, please change the code and docstrings for this function.

samples/javascript_nodejs/09.message-routing/bot.js
```diff
@@ -65,20 +65,20 @@ class MessageRoutingBot {
...
+
             const utterance = context.activity.text.trim().toLowerCase();
 
             // handle conversation interrupts first
             const interrupted = await this.isTurnInterrupted(dc, utterance);
-            if(interrupted) {
-                return;
+            let dialogResult;
+            if (interrupted) {
+                dialogResult = interrupted;
```
stevengum: The return value from `this.isTurnInterrupted()` is a boolean, why is the dialogResult being set to this?

samples/javascript_nodejs/09.message-routing/bot.js
```diff
@@ -65,20 +65,20 @@ class MessageRoutingBot {
...
+            } else {
+                dialogResult = await dc.continueDialog();
             }
 
-            // Continue the current dialog
```
stevengum: Please move this comment to above the `await dc.continueDialog();`

stevengum:
>      * @param {string} utterance - user's utterance

This wasn't in your changes, but can you explain that this argument has been normalized via the `.trim().toLowerCase()` calls? Otherwise there really isn't a point in passing in the "utterance" as it's just the `turnContext.activity.text`.

---
Refers to: samples/javascript_nodejs/09.message-routing/bot.js:119 in d9ff20d. [](commit_id = d9ff20d0466858fba85bd3ea808368625dedf05c, deletion_comment = False)

stevengum:
>         utterance = utterance.trim().toLowerCase();

And get rid of this line as it's redundant.

---
Refers to: samples/javascript_nodejs/09.message-routing/bot.js:123 in d9ff20d. [](commit_id = d9ff20d0466858fba85bd3ea808368625dedf05c, deletion_comment = False)

PR #727, opened 2018-10-18
Multilingual-conversations Node.js sample: minor fixes and improvements

carlosscastro:


samples/javascript_nodejs/17.multilingual-conversations/bot.js
```diff
@@ -67,8 +65,26 @@ class MultilingualBot {
...
+    // is different from the current language, then a change was indeed requested
+    return utterance !== currentLanguage;
+}
+
+/**
+ * Checks whether the utterance from the user is one of the 2 supported language codes in this sample
+ * @param {string} utterance the current turn utterance.
  */
-function isLanguageChangeRequested(utterance) {
+function isSupportedLanguageCode(utterance) {
```
stevengum: I don't understand this function, is it supposed to check if the the utterance is one of the supported languages? If so, the if statement should probably be something like this:

```js
// If the utterance matches one of the supported languages, the user is requesting a language change.
if ([ENGLISH_LANGUAGE, SPANISH_LANGUAGE].includes(utterance)) {
  return false;
}

return true; 
```
And the function should be renamed to something like _isNotSupportedLanguageCode_.

Also the utterance isn't cleaned up before this, so you should probably call `.trim().toLowerCase()` on it.

samples/javascript_nodejs/17.multilingual-conversations/translator-middleware.js
```diff
@@ -57,7 +57,7 @@ class TranslatorMiddleware {
      */
     async translate(text, to) {
         // Check to make sure "en" is not translated to "in"
-        if (text.toLowerCase() === ENGLISH_LANGUAGE) {
+        if (text.toLowerCase() === ENGLISH_LANGUAGE || text.toLowerCase() === SPANISH_LANGUAGE) {
```
stevengum: Does the comment above this line need to be updated?

samples/javascript_nodejs/17.multilingual-conversations/bot.js
carlosscastro: The function isSupportedLanguageCode checks if this is a supported language code. The name seems pretty self-explanatory but I'll look into reviewing it and the comments.

TBH I don't agree with your suggestion because of the following:

1- My proposed change is a boolean expression using already created constants and variables
2- [ENGLISH_LANGUAGE, SPANISH_LANGUAGE] requires creating a new array, reserve memory for it, assign it to the variable, and then add the 2 new elements
3- .includes means that we'll do a search in the array that we just created
4- return true / false seems repetitive when we could jsut do return [ENGLISH_LANGUAGE, SPANISH_LANGUAGE].includes(utterance);

If you feel very passionate about it or it is an idiomatic tihng I don't mind doing it this way, but in general programming this feels worst than what I proposed in many regards. 

I will add the cleanup, for sure!

samples/javascript_nodejs/17.multilingual-conversations/translator-middleware.js
carlosscastro: I can update it, sure it was just for generality

carlosscastro:
Thanks for the review @stevengum ! I'll send a new version shortly

samples/javascript_nodejs/17.multilingual-conversations/bot.js
carlosscastro: Also, regarding the isNotSupportedLanguageCode, in general its kind of an antipattern adding a "Not" to a logical method name, since it can further complicate things, I just prefer adding !condition rather than naming the method Notcondition. Hope that makes sense

stevengum: This whole discussion was because GitHub didn't show me the rest of the code, when I could have seen the entire function in CodeFlow.... Lesson learned!

PR #731, opened 2018-10-19
fix #593 by using Node.js model and generating reminders.lu in .netcore sample

### LINKED ISSUE #593: LUIS model is not the same for JS and C# samples
## Sample information
1. Sample type: samples
2. Sample language: **dotnetcore** and **nodejs**
3. Sample name: **12.nlp-with-luis**

## Describe the bug
The cognitive model used in the two versions of this sample are not quite the same.

## To Reproduce
1. Compare the JS file [reminders.json](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/12.nlp-with-luis/cognitiveModels/reminders.json) to the C# file [LUIS-Reminders.json](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/12.nlp-with-luis/CognitiveModels/LUIS-Reminders.json).
1. Note that one generates the intents: `Calendar.Add`, `Calendar.Find`, and `None`; while the other generates the intents: `Calendar-Add`, `Calendar-Find`, and `None`.

## Expected behavior
These two files should:
1. Have the same name, except for language-specific casing rules for files.
1. Have the same contents.

## Additional context
Any obvious logic or behavioral discrepancies between language-specific versions of a sample make them harder to use as the basis for documentation.

[bug]

stevengum:
Fixes #593


alundiak:
250 commits? WOW?! This PR is wrong for sure. Something not OK. !!!

PR #738, opened 2018-10-21
Introducing bot translation library

amr-elsehemy:
## Proposed Changes

- The sample shows how to use the library through Middleware to support multilingual interaction with bots in general and LUIS bots in particular.
- The main feature that differentiate this sample from the raw API is the ability to apply some customization to the raw translation using no-translate patterns interpreted by the translator library and/or custom vocab dictionary, 
   - example : [mon nom est l'etat] : the raw translation for this statement from "fr" to "en" should be "My name is the state", but using the patterns customization you can customize your output to be "My name is l'etat" which is the correct translation in this case 

## Testing

- Unit and integration tests are included for the translation library.


cleemullins:
@amr-elsehemy Can you please rebase this against master, at which point we can decide what to do here.

amr-elsehemy:
@cleemullins, since no review effort has started yet, I prefer to create a new clean PR 
#811 
and I'll abandon this one .

Issue #730, opened 2018-10-19
[netcore] FileNotFoundException when trying to load configuration file


dfavretto:
## Sample information
1. Sample type: \samples\
2. Sample language: dotnetcore

## Describe the bug
At the present time, the NETCore samples are loading `.bot` files without previously validating if they exist or if the file name is correct, leading to unhandled exceptions.

## To Reproduce
1. Go to sample `17.multilingual-bot` (netcore) which is missing a `.bot` configuration file of name `BotConfiguration.bot`
2. Follow Readme instructions to run the sample
3. When trying to get the configuration file, an error of type `System.IO.FileNotFoundException` is thrown with the message `Could not find file <PATH_TO_FILE>`.

## Expected behavior
The configuration file should be imported correctly; in case the file isn't present or its name isn't the expected, a user friendly error-message should be shown.

## Describe the solution you'd like
Alternatives:
1. Validate using `File.Exist(botFilePath)`.
2. Enclose the code block with `try-catch`, just like it's done in [NodeJS samples](https://github.com/Microsoft/BotBuilder-Samples/blob/70eb9862b0794c67583c01758bf387bed0c13f0e/samples/javascript_nodejs/10.prompt-validations/index.js#L31).

## Screenshots
Unhandled exception:
![image](https://user-images.githubusercontent.com/23645105/47218657-e53d2900-d382-11e8-92cd-fc1f990fa92a.png)

PR #765, opened 2018-10-23
Add FileExists check for .bot file to samples

nyxsys:
Fixes https://github.com/Microsoft/BotBuilder-Samples/issues/730

## Proposed Changes
I have added a `File.Exists` check for the `botFilePath` to the Echobot. I've tested this and it currently works and sends an error message to the browser, but I'd like to confirm that this implementation is what we'd like to go with before adding it to the rest of the C# samples. 

Additionally note that the sample bot `02.echo-with-counter` already includes a try-catch for if the file was read properly after the current check, most other bot samples do not include this try-catch block.

## Testing
Delete the .bot file from the sample and then attempt to run it.

### DIFF
--- a/samples/csharp_dotnetcore/02.echo-with-counter/Startup.cs
+++ b/samples/csharp_dotnetcore/02.echo-with-counter/Startup.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.IO;
 using System.Linq;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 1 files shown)


samples/csharp_dotnetcore/02.echo-with-counter/Startup.cs
```diff
@@ -62,7 +63,10 @@ public void ConfigureServices(IServiceCollection services)
 
                 var secretKey = Configuration.GetSection("botFileSecret")?.Value;
                 var botFilePath = Configuration.GetSection("botFilePath")?.Value;
-
+                if (!File.Exists(botFilePath))
+                {
+                    throw new FileNotFoundException($"The .bot configuration file was not found. botFilePath: {botFilePath}");
```
cleemullins: Looking at the code below, it appears if the botFilePath is not set (aka: set to null), then the code will default to ".\Botconfiguration.bot". This looks deliberate in the code below. 

I don't think this is a change that's needed. 

Now - a change that seems like a good idea would be to add the botfilePath (or default file name) to the exception message that's thrown on line 84.

nyxsys: I agree with adding the filepath to that location, but this check is for if the botFilePath (either the one thats set or the default one) does not actually refer to a file that exists, currently this throws a "could not find file" error that isn't caught (seen in screenshot attached below from issue #730), but the bot still won't work. 

It also appears that this particular sample is an exception with regard to error handling, as the logic around line 84 does not exist in most of the other sample bots, so this would cover the use case in those bots as well if its approved for putting everywhere.

https://user-images.githubusercontent.com/23645105/47218657-e53d2900-d382-11e8-92cd-fc1f990fa92a.png

PR #791, opened 2018-10-28
Samples c#: Update README.md

daltskin:
Updating the qna key is not necessary as it's provisioned as "sample-qna" by default.  Therefore line 46 doesn't need to be updated.

Fixes #

## Proposed Changes

  - 
  -
  -


alundiak:
I would add prefix `[samples c#]` in PR title. Coz my first thought was it's change in readme.md I have in my PR for generators :)

Issue #744, opened 2018-10-22
[webapi][02.echo with counter] Sample crashes on start due to duplicated port configuration


enzocanoo:
## Sample information
1. Sample type: **samples**
2. Sample language: **webapi**
3. Sample name: **02.echo-with-counter**

## Describe the bug
When running the sample, the following error gets thrown when the bot is loaded:
```
Could not find file 'C:\Users\username\Documents\Repos\BotBuilder-Samples\samples\csharp_webapi\13.Basic-Bot-Template\BasicBot.bot'. 
Description: An unhandled exception occurred during the execution of the current web request. Please review the stack trace for more information about the error and where it originated in the code. 

Exception Details: System.IO.FileNotFoundException: Could not find file 'C:\Users\username\Documents\Repos\BotBuilder-Samples\samples\csharp_webapi\13.Basic-Bot-Template\BasicBot.bot'.

Source File: C:\Users\username\Documents\Repos\BotBuilder-Samples\samples\csharp_webapi\13.Basic-Bot-Template\App_Start\BotConfig.cs    Line: 34
```

## To Reproduce
_Since there's no README in the sample._
1. Open the project with Visual Studio 2017.
2. Build the solution.
3. Run the project **EchoBot**.
4. See error when the bot ties to  load the `.bot` file.

## Expected behavior
Sample should run without any problems since there's no required services that should be deployed in order to work.

## Proposed solution
Change the endpoint configuration in the `.csproj` files to a unique port per sample. This change would require changing the `.bot` files endpoints too.

## Additional context
The `.csproj` files, seems to have the same endpoint configuration URL.
See [line 191](https://github.com/Microsoft/BotBuilder-Samples/blob/0492c45717fcb3033106572e8a1e55870414241f/samples/csharp_webapi/02.echo-with-counter/EchoBot.csproj#L191) of the `EchoBot.csproj` file. The same port is used by the BasicBot.csproj [here](https://github.com/Microsoft/BotBuilder-Samples/blob/0492c45717fcb3033106572e8a1e55870414241f/samples/csharp_webapi/13.Basic-Bot-Template/BasicBot.csproj#L241).
At the `EchoBot.csproj` file, the  `<WebProjectProperties>` is the same as the `BasicBot.csproj` file
```xml
[..]
<WebProjectProperties>
    <UseIIS>True</UseIIS>
    <AutoAssignPort>True</AutoAssignPort>
[...]

PR #799, opened 2018-10-29
[webapi] Modify port configuration

dfavretto:
Fixes https://github.com/Microsoft/BotBuilder-Samples/issues/744

## Proposed Changes
Port configuration was changed in each `.csproj`, and the change was reflected in `.bot` files.
This was done to avoid port duplication in the projects, leading to the overriding of the last served page.

## Testing
Running the sample `02.echo with counter` will not throw exception at start.


cleemullins:
@johnataylor @daveta Can you take a look? I don't quite understand why this is being done..

PR #808, opened 2018-10-30
Fix bot file path in welcome-users sample

ahmedlhanafy:
Fix bot file path in welcome-users sample


joseques:
Hi @ahmedlhanafy 
Thanks for the fix, I missed this change in the last PR renaming the `.bot` file to match the sample name!

PR #816, opened 2018-10-30
Use obj spread instead of Object.assign

leandrogrillo:
It's a chore, just :nail_care: 

## Proposed Changes
Instead of using object.assign, we can use obj spread in `activity` and simple object


cleemullins:
@leandrogrillo This looks reasonable to me, but I'm a C# developer and often the subtlies of JS are lost on me. :) 

I've added a JS dev to the reviewers list, and so we can hopefully get this merged in soon.

PR #805, opened 2018-10-29
Add fileExists check to all dotnet samples

nyxsys:
Fixes https://github.com/Microsoft/BotBuilder-Samples/issues/730

## Proposed Changes
Apply the change from previous PR https://github.com/Microsoft/BotBuilder-Samples/pull/765 to the other dotnet samples that include a bot file.

## Testing
Built and ran the samples this change was applied to.

PR #825, opened 2018-11-01
Johtaylo/custom storage

johnataylor:
New sample that demonstrates how to build custom storage solutions with differing policy.


samples/csharp_dotnetcore/42.scaleout/Startup.cs
```diff
@@ -63,10 +63,6 @@ public void ConfigureServices(IServiceCollection services)
             {
                 var secretKey = Configuration.GetSection("botFileSecret")?.Value;
                 var botFilePath = Configuration.GetSection("botFilePath")?.Value;
-                if (!File.Exists(botFilePath))
-                {
-                    throw new FileNotFoundException($"The .bot configuration file was not found. botFilePath: {botFilePath}");
-                }
```
dfavretto: This was introduced in #805; it seems that it shouldn't be removed.

Issue #812, opened 2018-10-30
Typescript Generator and Samples Lack Typing


ryanvolum:
TypeScript samples and the Yeoman generator lack typing. The generator uses require syntax, which implicitly types all imports as any: 
```
const { ActivityTypes } = require('botbuilder');
```
The right Types are not imported, so we end up with untyped code like: 
```
 async onTurn(turnContext) {
```
or
```
  constructor(conversationState) {
```

Any TypeScript developer will be immediately blocked - they'll have to backfill the types (if they can figure out how properties should be typed). Further, most TS developers use strict compiler options and ts lint rules, which would make samples/generators uncompilable. 

I would recommend developing TS samples with strict typing/linting, so that any TS developer can seamlessly get started with samples and templates.

sgellock:
fixed with PR #861

PR #861, opened 2018-11-06
Update Yeomen generated typescript

sgellock:
- use typescript syntax for echo generated template
- fix a see also link in basic bot generated template


sgellock:
Fixes #812

generators/generator-botbuilder/generators/app/templates/echo/bot.ts
```diff
@@ -1,41 +1,48 @@
...
+    private readonly conversationState: ConversationState;
+
+    /**
+     * 
+     * @param {ConversationState} conversation state object
+     */
+    constructor (conversationState: ConversationState) {
+        // Create a new state accessor property. 
+        // See https://aka.ms/about-bot-state-accessors to learn more about the bot state and state accessors.        
+        this.countAccessor = conversationState.createProperty(TURN_COUNTER_PROPERTY);
```
stevengum: This could be `this.countAccessor = conversationState.createProperty<number>(TURN_COUNTER_PROPERTY);`

generators/generator-botbuilder/generators/app/templates/echo/bot.ts
```diff
@@ -1,41 +1,48 @@
...
-      await this.countProperty.set(turnContext, count);
-    } else {
-      await turnContext.sendActivity(`[${turnContext.activity.type} event detected]`);
+    
+    private readonly countAccessor: StatePropertyAccessor<number>;
+    private readonly conversationState: ConversationState;
+
+    /**
+     * 
+     * @param {ConversationState} conversation state object
```
stevengum: The syntax highlighting on this line seems a bit odd; we might want to double check this docstring.

sgellock: this highlighting is the same as other docstrings in this file

stevengum: I looked into this a litle more, the docstrings here are for JS, not TS, we use [JavaDoc(?)](http://typedoc.org/guides/doccomments/) syntax, here's an example for [CardFactory.audioCard()](https://github.com/Microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/src/cardFactory.ts#L102-L116):

```ts
/**
 * Returns an attachment for an audio card.
  * @param title The cards title.
  * @param media Media URL's for the card.
  * @param buttons (Optional) set of buttons to include on the card.
  * @param other (Optional) additional properties to include on the card.
  */
public static audioCard(
  title: string,
  media: (MediaUrl | string)[],
  buttons?: (CardAction | string)[],
  other?: Partial<AudioCard>
): Attachment {
  return mediaCard(CardFactory.contentTypes.audioCard, title, media, buttons, other);
}
```

That said, TS does support [JSDoc](https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript), so if the current syntax is kept, the first token after the `@param {ConversationState} ` should be the parameter name, i.e.:

```ts
/**
 *
 * @param {ConversationState} conversationState <Brief conversation state description>
 */
```

PR #938, opened 2018-11-15
Rename proactive-messages.bot to BotConfiguration.bot

Kaiqb:
Proactive Messages sample throws exception because the .bot file name is not correct.

Renamed the .bot file to **BotConfiguration.bot** to match it with what is used in Startup.cs (see below):
`var botConfig = BotConfiguration.Load(botFilePath ?? @".\BotConfiguration.bot", secretKey);`
`var config = BotConfiguration.Load(@".\BotConfiguration.bot");`


cleemullins:
@Kaiqb We recently renamed all the .bot files to follow the "named after their project". 

In this case, it looks like the source code needs to be updated. I'll ping @JasonSowers who should fix.

JasonSowers:
fixed in #941

PR #957, opened 2018-11-19
CafeBot not dispatching to BookTable correctly.

### LINKED ISSUE #937: cafe-bot card input does not work, there's an unused method in the code, code differs from the NodeJS sample.
## Sample information
1. Sample type: samples
3. Sample name: 51.cafe-bot

## Describe the bug
Interacting with cards leads to a Null exception. The general functionality of the cafe-bot differs from the example in the NodeJS samples. 

## To Reproduce
1. Setup the sample.
2. Reach the WhatCanYouDo or WhoAreYou dialogs, make sure the cards display
3. Try to interact with the cards.

## Expected behavior
The card input should work and the bot should be able to extract information from cards. 

## Additional context
The code also includes the following method in the MainDispatcher.cs: 

```csharp
private async Task<DialogTurnResult> BeginWhatCanYouDoDialogAsync(DialogContext innerDc, OnTurnProperty onTurnProperty)
        {
            var context = innerDc.Context;

            // Handle case when user interacted with what can you do card.
            // What can you do card sends a custom data property with intent name, text value and possible entities.
            // See ../WhatCanYouDo/Resources/whatCanYouDoCard.json for card definition.
            var queryProperty = (onTurnProperty.Entities ?? new List<EntityProperty>()).Where(item => string.Compare(item.EntityName, QueryProperty) == 0);
            if (queryProperty.Count() > 0)
            {
                Dictionary<string, string> response;
                try
                {
                    response = JsonConvert.DeserializeObject<Dictionary<string, string>>(queryProperty.ElementAtOrDefault(0).Value as string);
                }
                catch
                {
                    await context.SendActivityAsync("Choose a query from the card drop down before you click 'Let's talk!'");
                    return new DialogTurnResult(DialogTurnStatus.Empty, null);
                }

                if (response.TryGetValue("text", out var text))
                {
                    context.Activity.Text = text;
                    await context.SendActivityAsync($"You said: '{context.Activity.Text}'.");
                }

                // Create a set a new onturn property
                await _onTurnAccessor.SetAsync(context, OnTurnProperty.FromCardInput(response));
                return await BeginDialogAsync(innerDc, response);
            }

            return await innerDc.BeginDialogAsync(WhatCanYouDo.Name);
        }
```
In the NodeJS version this code is called when the dispatcher hits the WhatCanYouDo intent. The dotnetcore sample simply calls the WhatCanYouDo dialog directly instead of calling the method which then calls the dialog. 
[bug]

daveta:
Fixes #937

## Proposed Changes
- Modify README.md to use `--noDecorate` in clone instructions.
- Modify code so user doesn't have to modify LUIS reference in code (--noDecorate).
- Fix BookTable dispatch so it runs BookTable correctly.
- Fix default bot file in startup (recently changed to `cafe-bot.bot`)
- Fix up storage issues associated with null SetAsync() which fail in ConcurrrentDirectionary.
- Fix JObject paths to reference LUIS results correctly.
- Remove SendAsync() with empty string.

## Testing
- Tested with latest msbot/az client with undecorated LUIS.

### DIFF
--- a/samples/csharp_dotnetcore/51.cafe-bot/CafeBot.cs
+++ b/samples/csharp_dotnetcore/51.cafe-bot/CafeBot.cs
@@ -32,14 +32,11 @@ namespace Microsoft.BotBuilderSamples
         public const string OnTurnPropertyName = "onTurnStateProperty";
         public const string DialogStateProperty = "dialogStateProperty";
 
-        // The name of the bot you deployed.
-        public static readonly string MsBotName = "cafe66";
-
         /// <summary>
         /// Key in the bot config (.bot file) for the LUIS instances.
         /// In the .bot file, multiple instances of LUIS can be configured.
         /// </summary>
-        public static readonly string LuisConfiguration = MsBotName + "_" + "cafeDispatchModel";
+        public static readonly string LuisConfiguration = "cafeDispatchModel";
 
         // Greeting Dialog ID.
         public static readonly string GreetingDialogId = "greetingDialog";
@@ -152,6 +149,7 @@ namespace Microsoft.BotBuilderSamples
                     {
                         await SendWelcomeMessageAsync(turnContext);
                     }
+
                     break;
                 default:
                     // Handle other activity types as needed.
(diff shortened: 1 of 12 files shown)


cleemullins:
@vishwacsena I don't have enough context here to have an opinion. Everything here looks reasonable to me, but if you could look and review, I think that would be best.

PR #967, opened 2018-11-26
make all onTurnError functions consistent

sgellock:
- use the load, clear, save pattern for samples that use state


sgellock:
if https://github.com/Microsoft/BotBuilder/issues/5175 gets done for 4.2, then this work will need to be redone to accommodate the new API.  For now, this PR makes all the samples consistent with the current API behavior.

PR #981, opened 2018-11-27
Updated npm scripts for TS Yeoman generators

### LINKED ISSUE #980: Watch script in generators and samples errors on first run
## Sample information
1. Sample type: TypeScript Samples/generators
2. Sample language: TypeScript
3. Sample name: TypeScript Samples/generators

## Describe the bug
The watch script in the TS samples/generators has a race condition - it tries to run compiled code before it's fully compiled. This means that on first run (before anything has ever been compiled), it throws a cryptic error. It then runs successfully once the code _has_ compiled, but the error is likely confusing to new devs. 

## To Reproduce
1. Run yeoman generator for echobot 
2. Run ```npm run watch```
3. Observe the error:

image

## Expected behavior
The bot should compile and run code. It should not throw errors on first compile. It should recompile whenever changes are made. 

## Additional context
Having contributed the concurrently code, I've got a fix in mind. I'll issue a PR to fix all samples/generators shortly. 

[bug]

ryanvolum:
Applies to #980 for the generators, but not the samples

## Proposed Changes
  - Resolves incorrect path to local commands (e.g. "node_modules/typescript/bin/tsc" should be "node_modules/.bin/tsc"). **Note**: There's a pretty good argument for using ```npx``` for this - it looks for local commands in .bin, and falls back on global commands if they don't exist. I chose to fix the path issue to get the watch issue resolved, but this should be revisited. 
  - Removes concurrently dependency given race condition
  - Uses nodemon typescript watch functionality (-e ts) to watch changes to TS files
  - Runs the start script each time a change is detected in any ts files in src (--watch ./src)

## Testing
I tested this by using ```npm link``` to run my local version of the generator. The watch script (and other fixed scripts) work as expected on the first run and on consecutive runs.

### DIFF
--- a/generators/generator-botbuilder/generators/app/templates/basic/package.json.ts
+++ b/generators/generator-botbuilder/generators/app/templates/basic/package.json.ts
@@ -6,10 +6,10 @@
     "license": "MIT",
     "main": "<%= npmMain %>",
     "scripts": {
-        "build": "node_modules/typescript/bin/tsc --build",
-        "start": "node_modules/typescript/bin/tsc --build && node ./lib/index.js",
-        "watch": "concurrently --kill-others \"node_modules/typescript/bin/tsc -w\" \"nodemon ./lib/index.js\"",
-        "lint": "node_modules/tslint/bin/tslint -c tslint.json 'src/**/*.ts'",
+        "build": "node_modules/.bin/tsc --build",
+        "start": "node_modules/.bin/tsc --build && node ./lib/index.js",
+        "watch": "node_modules/.bin/nodemon --watch ./src -e ts --exec \"npm run start\"",
+        "lint": "node_modules/.bin/tslint -c tslint.json 'src/**/*.ts'",
         "test": "echo \"Error: no test specified\" && exit 1"
     },
     "repository": {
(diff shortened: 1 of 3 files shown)


ryanvolum:
The `.bin` in the `node_modules` directory contains the "binaries" (not actually binaries, just following convention) of all of your node modules' cmd scripts. It effectively aggregates the contents of each package's `bin`. See the `tsc` command in .bin below:

image

So if we want to directly refer to a cmd it's less verbose to refer to `node_modules/.bin` than `node_modules/typescript/bin`.

Also, as they were, the npm scripts were throwing errors for me: 
image

These errors were resolved by referring to node_modules/.bin instead of referring to each package's bin, so as part of the `watch` fix, I changed paths appropriately.  

All that said, node will automatically look for local packages in .bin, and if they don't exist, it will then check your global directory, so `"build": "node_modules/.bin/tsc --build"` can just be `"build": "tsc --build",`. The only reason to refer to your bin directly (as we're doing here) is in case you want to make sure a developer is using your local package version, and not accidentally falling back on a global version of that package. This would only happen if they forgot to `npm install`. Given that these yeoman generators `npm install`s for you, I think we _should_  just use the command name instead of referring to its path. I'm happy to PR this in afterwards, but wanted this PR to be as small as necessary.

sgellock:
This was address in PR #1004

PR #983, opened 2018-11-27
included Bot Framework.zip in solution and VSIX by editing csproj

Zerryth:
Fixes https://github.com/Microsoft/botbuilder-dotnet/issues/1101

## Proposed Changes
  - included `Bot Framework.zip` (what contains the templates) in solution and VSIX by editing `csproj`

### DIFF
--- a/generators/vsix-vs-win/BotBuilderVSIX-V4/BotBuilderTemplatesVSIX.csproj
+++ b/generators/vsix-vs-win/BotBuilderVSIX-V4/BotBuilderTemplatesVSIX.csproj
@@ -71,6 +71,10 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       <IncludeInVSIX>true</IncludeInVSIX>
     </Content>
+    <Content Include="ProjectTemplates\Bot Framework.zip">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+      <IncludeInVSIX>true</IncludeInVSIX>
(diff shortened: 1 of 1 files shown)


cleemullins:
I don't think this is something we want to do. As part of the build, we have a build-step that is creating this zip on the fly, which means we don't have to directly check this in. 

You can see the Pre-Build event that's defined in the [.csproj](https://github.com/Microsoft/BotBuilder-Samples/blob/master/generators/vsix-vs-win/BotBuilderVSIX-V4/BotBuilderTemplatesVSIX.csproj) file:
```
 <PropertyGroup>
    <PreBuildEvent>powershell Compress-Archive -Path '$(SolutionDir)ProjectTemplates\Multi-project' -DestinationPath '$(ProjectDir)ProjectTemplates\Bot Framework.zip' -Force</PreBuildEvent>
  </PropertyGroup>
```

If you're doing something different, please swing by and let's chat so that we each understand what's going on.

EricDahlvang:
@cleemullins 

The .zip file is not checked in, but IS referenced by the .csproj  … if the .zip file is NOT referenced by .csproj with IncludeInVSIX = true, then it will NOT be in the resulting .vsix  


With @Zerryth changes, the pre-build step will create the .zip file, and it will be included in the resulting .vsix.

cleemullins:
Subsumed by #992 as part of the larger change set. Reviewed changes in person with @Zerryth to confirm.

PR #996, opened 2018-11-30
NodeJS: Update integration with Application Insights 

### LINKED ISSUE #976: NodeJS Application Insights middleware is missing event name values
## Sample information

1. Sample type: \samples\
2. Sample language: nodejs
3. Sample name: `20.qna-with-appinsights` and `21.luis-with-appinsights`

## Describe the bug
The included sample middleware for passing messages to Appllication Insights is missing some required field definitions.

In several places [like this one](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/20.qna-with-appinsights/middleware/myAppInsightsMiddleware.js), the code refers to variables like `this.botMsgReceivedEvent` which are undefined.  These fields should be defined, and should be set to useful event names as they are used to populate the event name in Application Insights.

This results in the events being ignored by Application Insights.

## To Reproduce

1. Configure one of the above samples per the directions, including app insights.
2. Run the sample, send and receive some messages.
3. Note that nothing appears in Application Insights.

## Expected behavior

After sending and receiving messages, application insights should at the very least show that _some events_ were received.

benbrown:
Fixes #976

## Proposed Changes
This updates the app insights samples to use the new botbuilder-applicationinsights telemetry client, which will improve the baseline information including in all events, and allow developers to take full advantage of the telemetry features being added in 4.2 

## Testing
In order to test this, you'll need to build it against [this branch](https://github.com/Microsoft/botbuilder-js/pull/662) of the Botbuilder libraries.  You should be able to use npm link or lerna to run against local packages until those make it into the daily builds.


samples/javascript_nodejs/20.qna-with-appinsights/index.js
```diff
@@ -4,14 +4,15 @@
...
+const { ApplicationInsightsTelemetryClient, ApplicationInsightsWebserverMiddleware } = require('botbuilder-applicationinsights');
 const { BotConfiguration } = require('botframework-config');
 const { QnAMakerBot } = require('./bot');
 const { MyAppInsightsMiddleware } = require('./middleware');
 
 // Read botFilePath and botFileSecret from .env file.
 // Note: Ensure you have a .env file and include botFilePath and botFileSecret.
 const ENV_FILE = path.join(__dirname, './.env');
-require('dotenv').config({ path: ENV_FILE });
+const env = require('dotenv').config({ path: ENV_FILE });
```
sgellock: I removed the 'const env =' code because it's not used in the sample.  the code assumes the process.env.

samples/javascript_nodejs/21.luis-with-appinsights/index.js
```diff
@@ -4,20 +4,21 @@
...
+const { ApplicationInsightsTelemetryClient, ApplicationInsightsWebserverMiddleware } = require('botbuilder-applicationinsights');
 const { BotConfiguration } = require('botframework-config');
 const { LuisBot } = require('./bot');
 const { MyAppInsightsMiddleware } = require('./middleware');
 
 // Read botFilePath and botFileSecret from .env file.
 // Note: Ensure you have a .env file and include botFilePath and botFileSecret.
 const ENV_FILE = path.join(__dirname, './.env');
-require('dotenv').config({ path: ENV_FILE });
+const env = require('dotenv').config({ path: ENV_FILE });
```
sgellock: I removed the 'const env =' code because it's not used in the sample. the code assumes the process.env.

benbrown:
@sgellock linter has been run, errors cleaned up.

PR #1004, opened 2018-12-03
make all onTurnHandlers use the same parameter name for context

### LINKED ISSUE #1001: onTurnError handler in Node 06.using-cards causes ReferenceError
## Sample information
1. Sample type: samples\
2. Sample language: nodejs
3. Sample name: 06.using-cards

## Describe the bug
The code for [onTurnError](https://github.com/Microsoft/BotBuilder-Samples/blob/3569dd5ff1c0f10e658131cf6a1d7fb18c4d1914/samples/javascript_nodejs/06.using-cards/index.js#L51-L62) (in the Node sample 06.using-cards) uses two different references to the TurnContext object, `turnContext` and `context`:

```js
// Catch-all for errors.
adapter.onTurnError = async (turnContext, error) => {
    // This check writes out errors to console log .vs. app insights.
    console.error(`\n [onTurnError]: ${ error }`);
    // Send a message to the user
    await turnContext.sendActivity(`Oops. Something went wrong!`);
    // Clear out state
    await conversationState.load(context);
    await conversationState.clear(context);
    // Save state changes.
};
```

If errors are emitted in the sample, they are not shown to the user as the user instead gets a ReferenceError and an UnhandledPromiseRejectionWarning.

[bug]

sgellock:
Fixes #1001 

## Proposed Changes
- fix some cut and paste errors where some routines were using context while others were using turnContext.
- made all samples be consistent

### DIFF
--- a/samples/javascript_nodejs/07.using-adaptive-cards/index.js
+++ b/samples/javascript_nodejs/07.using-adaptive-cards/index.js
@@ -49,11 +49,11 @@ const adapter = new BotFrameworkAdapter({
 });
 
 // Catch-all for errors.
-adapter.onTurnError = async (turnContext, error) => {
+adapter.onTurnError = async (context, error) => {
     // This check writes out errors to console log v.s. Application Insights.
     console.error(`\n [onTurnError]: ${ error }`);
     // Send a message to the user.
-    await turnContext.sendActivity(`Oops. Something went wrong!`);
+    await context.sendActivity(`Oops. Something went wrong!`);
     // Clear out state
     await conversationState.load(context);
     await conversationState.clear(context);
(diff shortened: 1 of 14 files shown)


samples/javascript_nodejs/07.using-adaptive-cards/index.js
```diff
@@ -49,11 +49,11 @@ const adapter = new BotFrameworkAdapter({
...
 // Catch-all for errors.
-adapter.onTurnError = async (turnContext, error) => {
+adapter.onTurnError = async (context, error) => {
     // This check writes out errors to console log v.s. Application Insights.
     console.error(`\n [onTurnError]: ${ error }`);
     // Send a message to the user.
-    await turnContext.sendActivity(`Oops. Something went wrong!`);
+    await context.sendActivity(`Oops. Something went wrong!`);
     // Clear out state
     await conversationState.load(context);
```
stevengum: This sample doesn't use any state, the `conversationState` logic can be removed.

PR #1024, opened 2018-12-05
Add MessageBack Card Action to Sample Bots

bwamie:
### NOTE: To be merged after MessageBack is implemented in WebChat and Emulator

Fixes # [Work Item](https://fuselabs.visualstudio.com/BotFramework/_queries/edit/17913)

## Proposed Changes
The purpose of this change is to add MessageBack card actions to bot samples. In sample 08 - suggested-actions, only ImBack ActionType was being used. Bot Framework now supports MessageBack ActionType. Also sample 06 - CardsBot was modified to add MessageBack card actions in Hero cards. Once MessageBack implementation is complete in the Emulator and WebChat bot developers will be encourage to use MessageBack instead of ImBack and PostBack. Therefore it is important that our samples demostrate how to use MessageBack ActionType.

### DIFF
--- a/samples/csharp_dotnetcore/08.suggested-actions/SuggestedActionsBot.cs
+++ b/samples/csharp_dotnetcore/08.suggested-actions/SuggestedActionsBot.cs
@@ -145,7 +145,9 @@ namespace Microsoft.BotBuilderSamples
         /// Creates and sends an activity with suggested actions to the user. When the user
         /// clicks one of the buttons the text value from the <see cref="CardAction"/> will be
         /// displayed in the channel just as if the user entered the text. There are multiple
-        /// <see cref="ActionTypes"/> that may be used for different situations.
+        /// <see cref="ActionTypes"/> that may be used for different situations. For this sample we have only used two.
+        /// Note that the button with MessageBack ActionType initializes Text instead of Value.
+        /// However, Value can be initialized as an additional property.
         /// </summary>
         /// <param name="turnContext">Provides the <see cref="ITurnContext"/> for the turn of the bot.</param>
         /// <param name="cancellationToken" >(Optional) A <see cref="CancellationToken"/> that can be used by other objects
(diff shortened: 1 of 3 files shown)


samples/csharp_dotnetcore/08.suggested-actions/SuggestedActionsBot.cs
```diff
@@ -145,7 +145,9 @@ private static string ProcessInput(string text)
         /// Creates and sends an activity with suggested actions to the user. When the user
         /// clicks one of the buttons the text value from the <see cref="CardAction"/> will be
         /// displayed in the channel just as if the user entered the text. There are multiple
-        /// <see cref="ActionTypes"/> that may be used for different situations.
+        /// <see cref="ActionTypes"/> that may be used for different situations. For this sample we have only used two.
+        /// Note that the button with MessageBack ActionType initializes Text instead of Value.
```
carlosscastro: If we introduce 2 Action types in the sample, we should explain the user the practical differences between both. 

Here is the right place for that. Why do we have multiple action types? If Message is the recommendation should the sample onlyuse message? Otherwise we need to explain when to use one or the other.

Fnally, the technical details such as 'Note that the button with MessageBack ActionType initializes Text instead of Value. However, Value can be initialized as an additional property.' Should probably go in line 161 at the code level/.

samples/csharp_dotnetcore/08.suggested-actions/SuggestedActionsBot.cs
```diff
@@ -158,9 +160,9 @@ private static async Task SendSuggestedActionsAsync(ITurnContext turnContext, Ca
             {
                 Actions = new List<CardAction>()
                 {
-                    new CardAction() { Title = "Red", Type = ActionTypes.ImBack, Value = "Red" },
-                    new CardAction() { Title = "Yellow", Type = ActionTypes.ImBack, Value = "Yellow" },
-                    new CardAction() { Title = "Blue", Type = ActionTypes.ImBack, Value = "Blue" },
+                    new CardAction() { Title = "Red", Type = ActionTypes.ImBack, Value = "Red: ImBack ActionType" },
+                    new CardAction() { Title = "Yellow", Type = ActionTypes.ImBack, Value = "Yellow: ImBack ActionType" },
```
carlosscastro: I just re-read the PR description and it says ` users will be encourage to use MessageBack instead of ImBack and PostBack.` Should we then only use MessageBack for 3 card actions here?

johnataylor:
this is very old now and we have updated since - reopen if you think this needs more work

PR #1028, opened 2018-12-06
[DO NOT MERGE] Luis with Application Insights v4.2

daveta:
Updates using new Application Insights support.
- Adds new TelemetryLoggerMiddleware
- Adds new TelemetryLuisRecognizer
   - New overload with DialogContext that will log DialogId
- Updated schema
- Bot Error handler now logs exceptions into IBotTelemetryClient for better troubleshooting.
- Application Insights registers for ILogger events.


samples/csharp_dotnetcore/21.luis-with-appinsights/Middleware/Telemetry/LuisTelemetryConstants.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.BotBuilderSamples
+{
+    /// <summary>
+    /// The Application Insights property names that we're logging.
+    /// </summary>
+    public static class LuisTelemetryConstants
```
ryanlengel: Can you switch to PascalCasing for all the values? Then it will be consistent with `TelemetryConstants.cs` and othe Solutions [LuisTelemetryConstants.cs](https://github.com/Microsoft/AI/blob/master/templates/Enterprise-Template/src/csharp/EnterpriseBotSample/Middleware/Telemetry/LuisTelemetryConstants.cs
)

daveta: Adding @benbrown.
Cool - Ben and I were having a discussion about the casing -- we're not inline with what appinsights uses, but will adopt your casing..

daveta: One thing to note: `TelemetryLuisRecognizer` having the overload with the `DialogContext`, it introduces a new dependency.  For non-dialog scenarios, kind of a drag.

daveta:
> @daveta can you validate if whether the SDK should be collecting the `ChannelId` property? I wasn't seeing it in my tests and added it to `BotMessageSend`/`BotMessageReceive` events so we can correlate with a conversation ID.

@ryanlengel : Yep, you should get channelId as a property.  Are you not seeing it?

PR #1047, opened 2018-12-06
[DO NOT MERGE][netcore] Implement Microsoft.Bot.PublishValidation package reference to netcore samples

dfavretto:
**NOTE:** Depends on the publication of Microsoft.Bot.PublishValidation package (Microsoft/botbuilder-dotnet#1214)

## Proposed Changes
  - Add Microsoft.Bot.PublishValidation package reference to all samples
  - Add properties for validating `LUIS` or `QnA Maker` keys in samples using those services
  - Remove use of `PrePublishValidate.cmd` and the file itself

## Testing
Run `dotnet publish` in any sample, and if the `.bot` file is correctly configured, the command won't fail. In case there's something not properly configured, a verbosely message will advert what is wrong.


cleemullins:
@dfavretto This is great work. For now, I suspect it's best to focus on just 1 or 2 samples and get them working as "Just Right". My vote would be for the EchoBotWithState sample (very simple, as it has no external dependencies) and perhaps a LUIS or QnA Sample. 

Once we've vetted that the solution and are confident we should bring it to the rest of the samples.

PR #1057, opened 2018-12-07
*DO NOT MERGE*: Application Insights 4.2 updates for Node JS

benbrown:
This is a repeat of #996 which was merged too early and rolled back by #1054 

This uses the new botbuilder-applicationinsights telemetry client which automatically correlates additional information along with the luis and qnamaker events.  

NOTE: This is pointing at the nightly build of botbuilder-applicationinsights.  To test you must first point npm to our nightly build feed.  

ALSO NOTE: We will have to update the package.json file to point to the public 4.2 version once it is published.


benbrown:
This has been updated with new README files, and the package.json files are now pointing at the public 4.2.0

Issue #1050, opened 2018-12-07
[C#] Multi-Lingual Sample fails with 4.2 Preview Builds


cleemullins:
As part of regression testing of the 4.2 release, @jwiley84 found the following bug:

```
Microsoft.BotBuilderSamples.MultiLingualBot:Error: Exception caught : Newtonsoft.Json.JsonSerializationException: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'Microsoft.BotBuilderSamples.Translation.Model.TranslatorResponse[]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.
To fix this error either change the JSON to a JSON array (e.g. [1,2,3]) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.
Path 'error', line 1, position 9.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)
   at Microsoft.BotBuilderSamples.Translation.MicrosoftTranslator.<TranslateAsync>d__6.MoveNext() in C:\Users\v-jewail\Desktop\regtest\BotBuilder-Samples\samples\csharp_dotnetcore\17.multilingual-bot\Translation\MicrosoftTranslator.cs:line 47

```

jwiley84:
...It helps if you remember to put the key in the appsettings file. That being said, that error message could have been MUCH clearer.

jwiley84:
A quick repro shows that if the key is blank, you get an appropriate error to update the key. If the key is INCORRECT (the sample comes with <Your translation key here> in the appsettings file), you get the above message.

PR #1062, opened 2018-12-08
Fix port number and improve error handling in MultiLingual sample bot (C#)

johnataylor:
- fixes port number in launch json
- throws on http error from translation services
- error handler send activity via adapter to explicitly avoid middleware
- minor change to logic so you can switch back from Spanish to English


johnataylor:
...and i left the translation auth key in the appsettings for now. the javascript version of this story actually uses process.env.

johnataylor:
fixes #1050

PR #1090, opened 2018-12-15
Remove superfluous comma in bot file and include hard-coded luis name…

### LINKED ISSUE #1089: [Node][21.luis-with-appinsights]: Sample has invalid botfile and doesn't run when corrected
## Sample information
1. Sample type: Sample
2. Sample language: nodejs
3. Sample name: 21.luis-with-appinsights/

## Describe the bug
On line 34 of the supplied bot file there is a comma which causes the JSON to fail validation and cause the bot to complain that there is no valid bot file.

If you fix that then you get a new error 
```
/home/andrew/projects/bot/index.js:59
    applicationId: luisConfig.appId,                             
TypeError: Cannot read property 'appId' of null
```

Which indicates that this line is returning null:
```
const luisConfig = botConfig.findServiceByNameOrId(LUIS_CONFIGURATION);
```

## To Reproduce
1. Install fresh project and run npm install
2. npm start and observe that there is a report about no botfile
3. remove the superfluous comma so that the botfile parses cleanly
4. npm start and observe that luisConfig is null

## Expected behavior
I would expect that luisConfig is populated with the values specified in the bot file

## Screenshots
Not applicable

## Additional context
$ nodejs --version
v11.3.0
$ npm --version
6.4.1
$ uname -r
4.15.0-39-generic

[bug]

andybeak:
Remove superfluous comma in bot file and include hard-coded luis name from index

Fixes #1089 

## Proposed Changes
Removing the comma makes the json parse cleanly so that the application does not complain about a missing botfile.

Including the name that the index.js file expects to be set in the botfile means that only user provided configuration needs to be added to it.  This makes it more clear what needs to be added.

## Testing
N/A

### DIFF
--- a/samples/javascript_nodejs/21.luis-with-appinsights/luis-with-appinsights.bot
+++ b/samples/javascript_nodejs/21.luis-with-appinsights/luis-with-appinsights.bot
@@ -12,7 +12,7 @@
         },        
         {
             "type": "luis",
-            "name": "",
+            "name": "luis-with-appinsights-luis",
             "appId": "",
             "authoringKey": "",
             "subscriptionKey": "",
@@ -31,7 +31,7 @@
             "applicationId": "",
             "apiKeys": {},
             "id": "4"
-        },
+        }
     ],
     "padlock": "",
     "version": "2.0"


jwiley84:
Hard-coding the LUIS name is not good practice, as it is easily changed in the index.js file, and will potentially confuse new bot framework users who are using the samples to learn how the framework operates. If the rest of the data in the bot file for the LUIS service is empty strings (""), then the name should be as well.

PR #1098, opened 2018-12-18
add conversationUpdate activity to browser sample

### LINKED ISSUE #945: [01.a.browser-echo] Does Not Receive conversationUpdate Activity.
## Sample information
1. Sample type: Sample
2. Sample language: JS
3. Sample name: 01.a.browser-echo

## Describe the bug
Can not greet incoming users using the current in-browser set-up. 

## Speculation 
there is no `conversationUpdate` activity sent, which is expected based on the samples for greeting the user. 

## To Reproduce
1. Clone Sample repo
2. Create Logic for Greeting user.
3. Nothing (Which is the error).

## Expected behavior
Ability to greet a user on bot start. 

Maybe related: #944 
[bug]

stevengum:
Fixes #945 

## Proposed Changes
  - The browser-echo sample now emits a conversationUpdate when the page is finished loading.
  - Changed the incoming and outgoing activities to have correct `from` and `recipient` fields.

issue945

### DIFF
--- a/samples/javascript_es6/01.a.browser-echo/src/app.js
+++ b/samples/javascript_es6/01.a.browser-echo/src/app.js
@@ -10,10 +10,15 @@ import { WebChatAdapter } from './webChatAdapter';
 // Create the custom WebChatAdapter.
 const webChatAdapter = new WebChatAdapter();
 
+// Create user and bot profiles.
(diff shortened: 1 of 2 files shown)


stevengum:
Pinging @Joll59

PR #1104, opened 2018-12-24
csharp_dotnetcore: Add echo bot sample, fixes to main README.md

sgellock:
## Proposed Changes
- add echo bot without state
- update main README.md to point to new sample
- removed incorrect :runner: emoji’s from main README.md


samples/csharp_dotnetcore/02.a.echo-bot/PREREQUISITES.md
```diff
@@ -0,0 +1,41 @@
...
+- Install the latest `botservice` extension for the Azure CLI tool.  Version 0.4.3 or higher.
+    1. Ensure any previous version of the `botservice` extension is removed.
+        ```bash
+        az extension remove -n botservice
+        ```
+    1. Install the latest `botservice` extension.
+        ```bash
+        az extension add -n botservice
+        ```
+    1. Verify you are running version 0.4.3 or higher of the `botservice` extension.
```
daveta: Nit: Numbering

samples/csharp_dotnetcore/02.a.echo-bot/Startup.cs
```diff
@@ -0,0 +1,82 @@
...
+using Microsoft.Bot.Connector.Authentication;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.BotBuilderSamples
+{
+    /// <summary>
+    /// The Startup class configures services and the request pipeline.
+    /// </summary>
+    public class Startup
```
daveta: >Startup [](start = 17, length = 7)

How far are we going with Application Insights?  if the answer is "Not in this sample", then we should at least show ILogger (ie, tracing) capabilities so developers can have some diagnostics if/when the bot fails so support has something to guide the user for exceptions.
For ILogger, you could add something like:
[https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/03.welcome-user/Startup.cs#L145](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/03.welcome-user/Startup.cs#L145)
[https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/03.welcome-user/Startup.cs#L81](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/03.welcome-user/Startup.cs#L81)
[https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/03.welcome-user/Startup.cs#L86](https://github.com/Microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/03.welcome-user/Startup.cs#L86)

If we want to double down on App Insights, the global exception handler could log into App Insights (the LUIS with App Insights samples would be a good one to copy).

daveta:
Question about App Insights and how early we want to weave that in.
Other than that looks awesome!

samples/csharp_dotnetcore/02.a.echo-bot/Program.cs
```diff
@@ -0,0 +1,38 @@
...
+                    // https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.1
+                    // logging.AddDebug();
+                    // logging.AddConsole();
+                })
+
+                // Logging Options.
+                // Consider using Application Insights for your logging and metrics needs.
+                // https://azure.microsoft.com/en-us/services/application-insights/
+                // .UseApplicationInsights()
+                .UseStartup<Startup>()
```
daveta: Kamran is working on some new docs for 4.2 that describe how to add instrumentation for Application Insights.  We should change this comment to reference that doc (when it gets published).

samples/csharp_dotnetcore/02.a.echo-bot/PREREQUISITES.md
sgellock: This is valid markdown.  the markdown render will number them correctly.  doing it this way allows you to move things, insert things, and never having to manually renumber them

samples/csharp_dotnetcore/02.a.echo-bot/Startup.cs
sgellock: makes sense.  I'll add ILogger support

sgellock: added ILogger support

PR #1127, opened 2019-01-05
Cleanup dotnet-template VS Code config files

drub0y:
* Remove duplicate entries in launch/tasks.json
 * Update the "build" task  in tasks.json to use the newer (2.x)
grouping mechanism and make it the default since
it's the only task there right now


drub0y:
@sgellock Here's the PR we were discussing offline earlier today to cleanup some of the VS code configuration issues in the `dotnet` templates.

drub0y:
@sgellock EmptyBot's launch.json was actually "ok", only the other ones had the duplicated entries and I didn't change anything else about the configurations.

sgellock:
Awesome.  Thanks Drew.

PR #1137, opened 2019-01-09
DO NOT MERGE: bot base class

johnataylor:
Take a look at the bot.js file now.

The switch on activity type (and the more involved unwinding of the conversation update) is moved into a base class. The application just extends that.

This is _very_ similar to the base controller class in the ASP MVC C# world.

### DIFF
--- a/samples/javascript_nodejs/30.bot-base-class/.env
+++ b/samples/javascript_nodejs/30.bot-base-class/.env
@@ -1,0 +1,2 @@
+botFilePath=botbaseclass.bot
+botFileSecret=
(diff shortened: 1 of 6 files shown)


johnataylor:
Incidentally Steven suggests merging the layered BotControllerBase and BotActivityControllerBase into a single class. Good suggestion. 

Then we derive and override onMessage to plug in the dialog containment class. We keep that separate because not every bot wants Dialogs although every bot wants to switch on activity type.

(Note all this switching is getting very close to routing. Perhaps we could thinking a harder about unifying thinks. The suggestion has been made that this might be a great way to handle channel types.)

PR #1146, opened 2019-01-11
Fix #1142 (cancellation confirmation in Cafe Bot)

### LINKED ISSUE #1142: [netcore][51.cafe-bot] The GetUserNamePrompt never cancels since the If statement that checks for the user response is incorrect
## Sample information
1. Sample type: **samples**
2. Sample language: **dotnetcore**
3. Sample name: **51.cafe-bot**

## Describe the bug
When the user tries to cancel out of the GetUserNamePrompt they get asked for confirmation, but the If statement that checks that response always lands on else (```return await base.ResumeDialogAsync(dc, reason, result);```), as such the prompt never cancels. 

## To Reproduce
1. Set up Cafe bot to be functional. 
2. Land on WhoAreYou dialog
3. Type "cancel" (so that LUIS lands on the Cancel Intent and then initiates the Cancel confirmation prompt).
4. The Yes answer fails to cancel the prompt, because the code never checks whether the response returned is True or False. 

## Expected behavior
The prompt should cancel appropriately when given the confirmation to do so by the user. 

## Screenshots
The behavior when the user responds with "Yes" to the Cancel prompt. 
image

image

## Additional context
Running this check ```if (result != null && (bool)result)``` gets the appropriate response and the prompt cancels (successfully reaches ```return await dc.CancelAllDialogsAsync();```) if the user clicked Yes. 

[bug]

v-kydela:
Fixes #1142 

While it was suggested I change `if (result == null)` to `if (result != null && (bool)result)`, I've used `result is bool` because that will return `false` if `result` is anything other than a `bool`, including `null`. If we only checked if it's `null`, the case `(bool)result` would throw an exception if `result` was a non-null object other than a `bool` (probably impossible in this case).

I've also modified another dialog where this pattern shows up, so they match.

### DIFF
--- a/samples/csharp_dotnetcore/51.cafe-bot/Dialogs/Shared/Prompts/GetLocationDateTimePartySizePrompt.cs
+++ b/samples/csharp_dotnetcore/51.cafe-bot/Dialogs/Shared/Prompts/GetLocationDateTimePartySizePrompt.cs
@@ -155,7 +155,7 @@ namespace Microsoft.BotBuilderSamples
 
         public async override Task<DialogTurnResult> ResumeDialogAsync(DialogContext dc, DialogReason reason, object result, CancellationToken cancellationToken = default(CancellationToken))
         {
-            if (result != null && result is bool && (bool)result != false)
+            if (result is bool && (bool)result)
             {
                 // User said yes to cancel prompt.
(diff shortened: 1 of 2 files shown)


v-kydela:
I can see this was merged but one of the tests failed. What was the test that failed? It seems I don't have access to the details.

PR #1176, opened 2019-01-29
Proposed: ASP.Net MVC Bot

johnataylor:
Controller 4 makes explicit use of the bot class - note how the adapter comes into play

Controller 5 demonstrates one of the possible (there are many) arrangements with DI

The bot class is decoupled from the controller and depends on the adapter.

This code also does the deserialization / serialization and mapping to HTTP explicitly - we should be able to hide that behind MVC smarts.


johnataylor:
Take a look at Controller 6.

- switching behavior is on a class called ActivityHandler
- Activity handler is derived from the **existing IBot interface**
- we add a new method to BotFrameworkAdapter see the code in BotFrameworkAdapterEx
(well not quite :-( ideally we'd add the method but in reality we need a slightly different method for WebApi and NetCore so we'll probably have to live with this "Ex" thing unless someone has a better idea)
- we define a new interface IBotFrameworkAdapter (again there would be two flavors of this - but really not the end of the world)
- note this new method makes use of the **existing BotCallbackHandler**
- the controller implementation is a single line of code
- and note the Dependency Injection - this factoring is _very friendly to dependency injection_ but absolutely doesn't require it
- this arrangement of passing in the Http objects is very close to the JavaScript world
- this manages to hide the awkward serialization
- the callbacks are strongly typed, in fact coupled with the whole override model this is _very Visual Studio intelli-sense friendly_
- I've added the collection and individual members added callbacks - so implementing "welcome" is a one liner but you can still easily get the collection if you like - I'm guess there are probably other protocol quirks we could capture like this - needs a little channel specific knowledge
- this is very plug and play - if you don't like the ActivityHandler you don't have to use it, _because this reuse the existing IBot interface moving existing bots across to this is extremely easy_, and development can be phased: fisrt you take your existing IBot and plug it in, then you derive from ActivityHandler and you can drop some of your existing code

samples/csharp_dotnetcore/30.asp-mvc-bot/MyActivityHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Bot.Builder;
+using Microsoft.Bot.Schema;
+
+namespace Asp_Mvc_Bot
+{
+    public class MyActivityHandler : ActivityHandler
+    {
+        protected override async Task OnMemberAddedAsync(ChannelAccount account, ITurnContext<IConversationUpdateActivity> turnContext, CancellationToken cancellationToken)
+        {
+            await turnContext.SendActivityAsync(MessageFactory.Text($"Welcome {account.Name}"));
```
benbrown: this makes me so happy.

cleemullins:
This work corresponds to DCR: https://github.com/Microsoft/BotBuilder/issues/5285

PR #1181, opened 2019-01-29
Generating a consistent story to BotServices between LUIS and QnAMaker Samples

mahomedalid:
Fixes https://github.com/MicrosoftDocs/bot-docs/issues/777

## Proposed Changes
LUIS uses BotServices with:

    A constructor that takes BotConfiguration and initializes a Dictionary

QnAMaker uses BotServices:

    With a static method that initializes the QnAMaker instances on a Dictionary
    Then the result is being passed to BotServices just as a container

It's a little confusing if you are working on a project with both LUIS and QnAMaker to reconcile both approaches on the same base code. Additionally, the BotServices abstraction seems like a recommended approach, which I don't know if it's the case.

### DIFF
--- a/samples/csharp_dotnetcore/11.qnamaker/QnAMaker/BotServices.cs
+++ b/samples/csharp_dotnetcore/11.qnamaker/QnAMaker/BotServices.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.Bot.Builder.AI.QnA;
+using Microsoft.Bot.Configuration;
 
 namespace Microsoft.BotBuilderSamples
 {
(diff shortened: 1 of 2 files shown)


samples/csharp_dotnetcore/11.qnamaker/QnAMaker/BotServices.cs
```diff
@@ -19,10 +20,51 @@ public class BotServices
...
-            QnAServices = qnaServices ?? throw new ArgumentNullException(nameof(qnaServices));
+            foreach (var service in botConfiguration.Services)
+            {
+                switch (service.Type)
+                {
+                    case ServiceTypes.QnA:
+                        {
+                            // Create a QnA Maker that is initialized and suitable for passing
+                            // into the IBot-derived class (QnABot).
+                            var qna = (QnAMakerService)service;
```
cleemullins: This should be:
`var qna = service as QnaMakerService; `
The cast as you have it now will throw, rather than return null.

samples/csharp_dotnetcore/11.qnamaker/QnAMaker/BotServices.cs
```diff
@@ -19,10 +20,51 @@ public class BotServices
         /// <summary>
         /// Initializes a new instance of the <see cref="BotServices"/> class.
         /// </summary>
-        /// <param name="qnaServices">A dictionary of named <see cref="QnAMaker"/> instances for usage within the bot.</param>
-        public BotServices(Dictionary<string, QnAMaker> qnaServices)
+        /// <param name="botConfiguration">A dictionary of named <see cref="BotConfiguration"/> instances for usage within the bot.</param>
+        public BotServices(BotConfiguration botConfiguration)
         {
-            QnAServices = qnaServices ?? throw new ArgumentNullException(nameof(qnaServices));
+            foreach (var service in botConfiguration.Services)
```
cleemullins: As an aside, we just add a bunch of these validations into the 4.3 (soon!) release.

samples/csharp_dotnetcore/11.qnamaker/QnAMaker/BotServices.cs
cleemullins: (I see that's from the old code, but... a bug is a bug is a bug... ) :)

mahomedalid: Indeed, fixing it.

samples/csharp_dotnetcore/11.qnamaker/QnAMaker/BotServices.cs
mahomedalid: Looking forward to it!

PR #1202, opened 2019-02-04
DO NOT MERGE: Demo of Typescript ActivityHandler class

benbrown:
This is an initial Javascript port of #1176

# Introducing ActivityHandler

This introduces a new class called ActivityHandler. ActivityHandlers have a `.run()` method that accepts the incoming turnContext created by `adapter.processActivity`.  The ActivityHandler then emits events based on the activity type and other characteristics.  These events can be handled by developer-defined functions.

The intention is to provide a flexible base class for bots.  This should make it easier for developers to handle specific types of activities, and easier to understand the scope of activities a bot might receive. All activity types that might arrive via the Azure Bot Service are exposed as events, along with a few specialized events to - for example - distinguish whether a conversationUpdates represents member added versus members removed.

Here's a simple example, showing an "Echo bot" that replies to any incoming message with an echo. 

```
const bot = new ActivityHandler();

server.post('/api/messages', (req, res) => {
    adapter.processActivity(req, res, async (context) => {
        // Route to main dialog.
        await bot.run(context);
    });
});


bot.onMessage(async (context, next) => { 
// do stuff
await context.sendActivity(`Echo: ${ context.activity.text }`);

// proceed with further processing
await next();
});

bot.onMembersAdded(async(context, next) => {
// send welcome messages to people listed in context.activity.membersAdded
await next();
});
```

## Events

For all incoming activities, a `Turn` event will fire. 

```
bot.onTurn(async (context, next) => { 
// this method receives every single incoming activity
await next();
});
```

Then, the type-specific event will fire. For example, ConversationUpdate:

```
bot.onConversationUpdate(async(context, next) => {
// this is a conversation update of some sort
await next();
```

Then, any sub-events will fire. For example, MembersAdded, which is a specialized version of ConversationUpdate:

```
bot.onMembersAdded(async(context, next) => {
// handle context.activity.membersAdded
await next();
});
```

Finally, a special `Dialog` event will fire that can be used to process Dialog activities.  (see below)

## Prevent Propagation of event

At any point in the processing chain, a handler may omit a call to `await next()` to prevent the event from continuing to be processed by other handlers. 

This pattern can be used to capture interruptions or handle special events before they reach other handlers.

## Using with Dialogs

In order to use an ActivityHandler along with the Dialogs system, use the `onDialog()` method, which fires at the very end of the processing chain. 

```
bot.onDialog(async (context, next) => {

    if (context.activity.type === ActivityTypes.Message) {
        const dialogContext = await dialogSet.createContext(context);
        const results = await dialogContext.continueDialog();

        // fallback behavior is to run welcome dialog
        if (results.status === DialogTurnStatus.empty) {
             await dialogContext.beginDialog('hello');
        }

    }

    await next();

});
```


benbrown:
Here's the related PR to add the ActivityHandler class directly into the SDK:
https://github.com/Microsoft/botbuilder-js/pull/757

Issue #1203, opened 2019-02-05
Deployed sample doesn't work on Azure


nstejaswini:
## Sample information
1. Sample language: dotnetcore 
2. Sample name: 03. welcome-user

## Describe the bug
The sample doesn't work on webchat in Azure after deploying it using the msbot clone command in readme.
## To Reproduce
1. Make the sample work locally.
2. Use msbot clone command (after obtaining the appId and appSecret) as given in readme to deploy the sample on Azure.
3. Test the deployed sample on Azure in webchat.
     - It doesn't work on webchat. If I try to send some message, it says 'can't send'.

## Expected behaviour
The sample should work according the readme.

jwiley84:
I'll have to check the README, but first: after you deployed, did you go into the deployed bot's .bot file and update it with a 'production' endpoint?

mdrichardson:
@nstejaswini 

I was able to follow the steps in the Readme and get this working, successfully.

My steps:

1. Clone the repo
2. Open the sample in Visual Studio
3. Press F5
4. Open in Emulator and tested successfully
5. Create a new app in the App Registration Portal
6. Generated a New Password and copied the AppId and Password
7. Used `msbot clone services` command from the bot's README, with the id/pass from above. This auto-generated the production endpoints in the `.bot` file.
8. Tested both production endpoints in Emulator (for some reason there's two)--success for both
9. Tested the app in Azure with "Test in Web Chat" -- success

![image](https://user-images.githubusercontent.com/40401643/52297514-c1075700-2935-11e9-8076-a5f95e84e3f5.png)


Was your deployment successful, or did it produce errors? If you can provide you `appId`, I can do additional research and look into this further.

nstejaswini:
I tried it again today twice. It worked for the second time. Last time I tried deploying two to three samples on Azure using the msbot command given in readme file, but the same error was coming. I don't know why that happened.  
It is not given in Readme that, before using the msbot clone command, path to sample (cd botBuilder/samples/…) should be given.

mdrichardson:
I looked into the appId you provided and I think there's something wrong with your app service or deployment. There's a couple things you can try from Azure Portal > Your App Service:

1. Try restarting the app service
2. Open the App Service Editor and verify that the `.bot` file that got deployed has a production endpoint, then check `appsettings.json` there and verify that it's using the `.bot` file that has the production endpoint. You may also want to click "Output" and then "Run" and see if it shows any errors:

[...]

PR #1231, opened 2019-02-13
Added note about matching LUIS key and config names for deployed bot

mdrichardson:
Fixes https://github.com/Microsoft/BotBuilder-Samples/issues/1203

## Proposed Changes

After `msbot clone services ...` runs, it often renames the LUIS service in the `.bot` file. I just added note about making sure to match the LUIS key and config names for the deployed bot to the relevant LUIS bot samples.


mdrichardson:
No longer matters.

PR #1239, opened 2019-02-14
[Various Samples] Added quotes around appSecret and appId for msbot cmd

### LINKED ISSUE #1233: Syntax error of msbot clone command
## Sample information
1. Sample language: dotnetcore 
2. Sample name: 14.nlp-with-dispatch

## Describe the bug
	Error while running  ‘msbot clone’ command. It says “The syntax of the command is incorrect” or “was unexpected this time”.
## To Reproduce
         1.Install the required tools.
         2.Run msbot clone command.
	- When I give the command with all the required parameters it shows errors like “The syntax of the command is incorrect” or “was unexpected this time”.
	When I give the appId and appSecret values in between “ “, then it doesn’t show any error.

## Expected behaviour
The sample should work according the readme.

mdrichardson:
Fixes #1233 

## Proposed Changes
Added `""` around `appId` and `appSecret` for `msbot clone services ...` command in README.md for applicable samples. This is a rare issue, but seems to pop up occasionally due in large part to whether `appSecret` contains special characters and in small part to the console the user uses.

### DIFF
--- a/samples/csharp_dotnetcore/02.b.echo-with-counter/README.md
+++ b/samples/csharp_dotnetcore/02.b.echo-with-counter/README.md
@@ -43,7 +43,7 @@ You can use the [MSBot](https://github.com/microsoft/botbuilder-tools) Bot Build
 
 To clone this bot, run
 ```bash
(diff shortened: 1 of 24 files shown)


mdrichardson:
No longer matters.

PR #1275, opened 2019-02-28
4.3: multi turn prompt bot

johnataylor:
- uses mvc api controller
- dependency inject all the things
- add ActivityHandler
- add root dialog
- add custom prompt validator


samples/csharp_dotnetcore/05.multi-turn-prompt/AdapterWithLogging.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Bot.Builder;
+using Microsoft.Bot.Builder.Integration.AspNet.Core;
+using Microsoft.Bot.Connector.Authentication;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.BotBuilderSamples
+{
+    public class AdapterWithLogging : BotFrameworkHttpAdapter
```
yochay: Maybe something like AdapterWithErrorHandling

samples/csharp_dotnetcore/05.multi-turn-prompt/Bots/DialogBasedBot.cs
```diff
@@ -0,0 +1,45 @@
...
+            _dialog = dialog;
+            _logger = logger;
+        }
+
+        protected override async Task OnMessageActivityAsync(ITurnContext<IMessageActivity> turnContext, CancellationToken cancellationToken)
+        {
+            var dialogSet = new DialogSet(_conversationState.CreateProperty<DialogState>("DialogState"));
+            dialogSet.Add(_dialog);
+
+            var dialogContext = await dialogSet.CreateContextAsync(turnContext, cancellationToken);
```
yochay: @johnataylor  can you bring the 'Run' method from sample 42 (or 30?)

johnataylor: We discussed just this @Kaiqb was in favor of keeping things flat. Perhaps we can iterate a little on this. The big step here is to move all the dialog logic into a ComponentDialog and inject the resulting class. That is what 42 does too. However, 42 also uses a much more straight forward state model. We tried that approach but then found we wanted to use the UserState for parity with the existing sample.

samples/csharp_dotnetcore/05.multi-turn-prompt/AdapterWithLogging.cs
johnataylor: OK makes sense. We just copied this class from the other project. I'l rename them together.

samples/csharp_dotnetcore/05.multi-turn-prompt/DialogExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using Microsoft.Bot.Schema;
+
+namespace Microsoft.BotBuilderSamples
+{
+    public static class DialogExtensions
+    {
+        public static async Task Run(this Dialog dialog, ITurnContext<IMessageActivity> turnContext, ConversationState conversationState, CancellationToken cancellationToken)
+        {
+            var dialogSet = new DialogSet(conversationState.CreateProperty<DialogState>("DialogState"));
+            dialogSet.Add(dialog);
```
JonathanFingold: This is kind of cool, but requires the dialog set to be constructed the same way every time. If the dev for some mysterious reason decides to add logic to choose between two different dialogs here, things will not go well.

samples/csharp_dotnetcore/01.console-echo/Console-EchoBot.csproj
```diff
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.0</TargetFramework>
-    <CodeAnalysisRuleSet>BotBuilder.ruleset</CodeAnalysisRuleSet>
```
johnataylor: this was not intentional????

johnataylor: we're going to clean up the staging when we have more PRs in there

PR #1276, opened 2019-02-28
v-stkanb/update-msgraph-sample

stevkan:
## Proposed Changes
Updated simple-graph-client.js to correctly access a user's AAD profile photo via an MS Graph API call and the handling of the data, once received, in oauth-helpers.js.


benbrown:
This is fixed in the `samples-work-in-progress` branch

PR #1327, opened 2019-03-13
Updates to BasicBot example for JS

benbrown:
## Proposed Changes

* Restructured the activity handler class and renamed it to "BotRunner" placed it in the components/ folder by itself.
* Added a method to easily load in "feature" modules containing handlers functions to bind to the activity handler
* Moved the "welcome" functionality out of the class and into into features/welcome.js
* Added a few additional sample "features" useful for debugging
* Added a ton of comments
* Rewrote a lot of the actual dialog text to be clearer to the user
* Added a fallback message if LUIS is not enabled
* Fixed some bugs


stevengum:
Closing as the bug fixes were taken separately

PR #13, opened 2016-09-30
[CreateNewConversation] Handling API Operation error when Channel does not support it

pcostantini:
b7d14906-8708-11e6-86e7-91bd2831e3f6

PR #1332, opened 2019-03-14
[MVC] Move 'BookingDetails' to 'Models' folder in Sample #13 - Basic Bot

GasparAcevedoZainSouthworks:
## Proposed Changes
- Create a folder named `Models` to contain the model classes

**Note:** We created this PR against `samples-work-in-progress` because here is where we reviewed the sample. On the other hand, we noticed these changes are also merged in `4.3/js/core` branch.
Let us know if we should rebase this PR.

## Testing
This modification does not change the code logic


johnataylor:
This is not a bad idea at all. Consistent with the mvc world. However, we are using this particular sample as a template for the template in teh generator. I'm not sure we want to take this change at this point. And then we will probably kill this basic-bot folder altogether.

benbrown:
We'll do this when we make some of the other updates to the samples.

PR #1344, opened 2019-03-19
Remove extra JS samples

Kaiqb:
To make nodejs samples in sync with C# samples.


samples/javascript_nodejs/13.basic-bot/dialogs/luisHelper.js
```diff
@@ -17,37 +16,38 @@ class LuisHelper {
...
+                endpoint: `https://${ process.env.LuisAPIHostName }`
             }, {}, true);
 
             const recognizerResult = await recognizer.recognize(context);
 
             const intent = LuisRecognizer.topIntent(recognizerResult);
-            const score = recognizerResult.luisResult.topScoringIntent.score;
-            
+
+            bookingDetails.intent = intent;
```
stevengum: What's this for? I don't think we use the intent in this sample...

Kaiqb: Not sure. This code is based on the changes that were done to Ben's update. I'm only removing samples and adding folder + readmes for additional samples we are migrating from another repo in this PR. You might want to check with Ben/Scott.

PR #1348, opened 2019-03-21
Fix dialog state not saving in C# sample 18: Authentication Bot

v-kydela:
This PR makes the C# sample closer to the Node version


johnataylor:
Thanks we got found this one last week and fixed it in the samples-work-in-progress branch.

Incidentally the waterfall was also incorrectly written and would have failed in some circumstances. Basically the prompt - and even the oauth prompt - must be the last statement executed in a waterfall step.

johnataylor:
So if we want to keep this sample around (for example in an archive of master) then we should take this fix and also fix the waterfall.

Alternatively we can drop the sample and just point people at the new much cleaned up and debugged sample.

cleemullins:
Closing, per @johnataylor comments.

PR #1354, opened 2019-03-25
Added missing await to context.sendActivity in onTurnError

stevkan:
## Proposed Changes

  - The "context.sendActivity" located in index.ts, line 61, in adapter.onTurnError is missing an await. Without, the bot fails with an ugly error.


sgellock:
@stevkan thanks for this change.  I've taken your fix and also applied your fix to the yeoman template generated code as well.  You can see this in the samples-work-in-progress branch until that gets merged into master.

PR #1377, opened 2019-04-03
Sets dialogId for UserProfileDialog to nameof(UserProfileDialog)

### LINKED ISSUE #1349: Multi-turn-prompt still uses "root" as its ID
## Sample information
1. Sample type: [\samples\]
2. Sample language: [dotnetcore]
3. Sample name: 05.multi-turn-prompt
4. Branch: **samples-work-in-progress**

## Describe the bug
The component dialog hard-codes its ID as `root`.

## To Reproduce
1. Look at https://github.com/Microsoft/BotBuilder-Samples/blob/7695212c2eb3dc22df7c86355407494626f28f13/samples/csharp_dotnetcore/05.multi-turn-prompt/Dialogs/UserProfileDialog.cs#L17-L18.
1. Error. John agreed this was not the right way to do this and said he'd change this to:
   ```csharp
        public UserProfileDialog(UserState userState)
            : base(nameof(UserProfileDialog))
   ```

## Expected behavior
Using a fixed, hard-coded string for the ID should not be used in samples, as this feels like a "bad habit" that could easily cause headaches and errors when reusing the code.

## Additional context
This is causing friction in the documentation that references this sample.

[bug]

gabog:
Fixes #1349 (see issue for more details)

### DIFF
--- a/samples/csharp_dotnetcore/05.multi-turn-prompt/Dialogs/UserProfileDialog.cs
+++ b/samples/csharp_dotnetcore/05.multi-turn-prompt/Dialogs/UserProfileDialog.cs
@@ -12,10 +12,10 @@ namespace Microsoft.BotBuilderSamples
 {
     public class UserProfileDialog : ComponentDialog
     {
-        private IStatePropertyAccessor<UserProfile> _userProfileAccessor;
+        private readonly IStatePropertyAccessor<UserProfile> _userProfileAccessor;
 
         public UserProfileDialog(UserState userState)
-            : base("root")
+            : base(nameof(UserProfileDialog))
         {
             _userProfileAccessor = userState.CreateProperty<UserProfile>("UserProfile");
 


gabog:
@JonathanFingold, just for future consideration, what would you use instead of nameof? 

DialogId is used to key off the dialogs collection, and it is also used in telemetry logging and help us map a log entry to the dialog that originated it. 

nameoff() helps us avoid name collission and makes it easier to refactor and maintain the code if the type gets renamed (this is harder if that's a hardcoded string).

The only issue that I see is that nameof doesn't include the namespace for the class and you may potentially have collisions if you create a dialog with the same name somewhere else. 

An alternative would be to use **typeof(*ClassName*).FullName** that returns a namespaced name for the type.

FYI @johnataylor.

JonathanFingold:
My suggestion would be to force the dev to assign the ID from the calling code, so something like:
```csharp
        public UserProfileDialog(string id, UserState userState)
            : base(id)
```

That way, if for some reason they want two of these dialogs in the same set, they can do that, each with a different ID.

PR #1390, opened 2019-04-09
[Typescript Samples] Make webConfigPrep work with Azure-generated web.config

### LINKED ISSUE #1255: Deployment script for typescript bot generator doesn't work
## Sample information
1. Sample type:` \generators`
2. Sample language: node, typescript
3. Sample name: `generators/generator-botbuilder/generators/app/templates/echo/deploymentScripts`

## Describe the bug
The ` webConfigPrep.js` file provided by the yeoman generate for the typescript bot is wrong (at least it didn’t work for me using Deployment Centre -> Azure Pipelines for CI/CD). It was looking to replace `url=index.js` but the web.config file generated for me by Azure was using `server.js`. You also need to update all instances pointing at `server.js`, not just url=

So I went back and updated the script to this which worked for me:

```
if (fs.existsSync(path.resolve(WEB_CONFIG_FILE))) {
    replace({
        regex: "server.js",
        replacement: "lib/index.js",
        paths: ['./web.config'],
        recursive: false,
        silent: true,
    })
}
```

[bug]

mdrichardson:
Fixes #1255

## Proposed Changes

When Kudu runs the start command, it generates a `web.config` that doesn't match the one that `az  bot prepare-deploy` generates. This ensures that they match by:

* Replace all references to `server.js` to `index.js`
* Replace `url="index.js"` with `url="lib/index.js"`
* Comment out `<iisnode watchedFiles="web.config;\*\.js" debuggingEnabled="false" \/>` if not already done

Note: I put this in the `samples-work-in-progress` repo because I know we're making big changes to the samples. I can add to the master repo, if requested.

## Testing

Before:

image

image

image

After:

image

image

image
^^ This is what success looks like


mdrichardson:
Complete. Again, this is a push to the `samples-work-in-progress` branch. Let me know if you want me to update the ones on `master`, too.

stevengum:
@mdrichardson please merge the latest commits in the branch into your clone; we took out the TS QnA Maker sample today. 

The only potential hiccup I can think of is that if the user has a simple JS web.config when they push their code up, is the new script going to still work with a web.config retrieved from `az bot prepare-deploy --lang Node`?

I'm not super concerned about this because our new stance is to just have users build their code before deployment and have all necessary configurations ready.

mdrichardson:
@stevengum I'll merge and then do some testing of `az bot prepare-deploy --lang Node`. Should be fine, but I'll make sure.

mdrichardson:
@stevengum I've merged with `samples-work-in-progress`. 

I also ran `az bot prepare-deploy --lang Node` and diffed against it after then running

`node deploymentScripts/webConfigPrep`

Other than changing `server.js` to `index.js` in a comment, it changed `url="index.js"` to `url="lib/index.js"`, as it should. No other diffs. So yes, this should work with both:

1) User runs `az bot prepare-deploy --lang Node` and generates a `web.config` on their own
2) User doesn't upload a `web.config` and Kudu generates one for them

mdrichardson:
@stevengum We can close this now, right, since `az prepare-publish` takes care of this? Although regarding my last comment, I *think* if we create that `index.js` file at the root, Kudu will generate a working `web.config` for us, which I *think* allows us to get rid of the `prepare-publish` step entirely?.

stevengum:
Re: changing `tsconfig.json` to output the `index.js` at the root level; it's a work around, however, I think that if you're going to be making a TypeScript project, generally the transpiled JS is all going to the same spot. 

We don't want to get into the business of telling people how to build their code just to get it to work on Azure. With the current solution in place, it works specifically with our samples, but it's easy enough to change the output `web.config` to match what a user might want.

PR #1393, opened 2019-04-09
Fix issues with proactive bot sample

EricDahlvang:
- Enable the proactive messaging to function even when there is not appid/password (emulator)
 - Echo user responses (instead of nothing)
 - Provide a welcome message in OnMembersAddedAsync, with link to notify controller
 - Respond with text of who was proactively messaged in the Notify Controller


EricDahlvang:
The node proactive sample still needs to be simplified like this one.

johnataylor:
@Kaiqb can you take a quick look. I think this is all ok. It does add a couple of things but makes the sample work better out of the gate. I don't think I breaks the spirit of keeping things simple.

EricDahlvang:
@Kaiqb  Feedback addressed: https://github.com/Microsoft/BotBuilder-Samples/pull/1393/commits/5a8f857233263c010a28fb26f6313bd1cd7d56ce

PR #1404, opened 2019-04-15
added QnA Dialog

johnataylor:


experimental/qnamaker-prompting/csharp_dotnetcore/Helpers/CardHelper.cs
```diff
@@ -0,0 +1,49 @@
...
+        public static Activity GetHeroCard(string cardTitle, QnAPrompts[] prompts)
+        {
+            var chatActivity = Activity.CreateMessageActivity();
+            var buttons = new List<CardAction>();
+
+            var sortedPrompts = prompts.OrderBy(r => r.DisplayOrder);
+            foreach (var prompt in sortedPrompts)
+            {
+                buttons.Add(
+                    new CardAction()
```
feifeiJun: Hey, I clone your porject and debug locally, it indicates that, the follow-up prompot are made to buttons, when click the button, the bot will send the message on this button to bot service. That' s a great design. 
I have a request on QnA follow-up prompt, **Could you please help to provide some ideas or suggestions?** In my scenario, those buttons are created by the follow up prompts from QnA, and the prompts contain some information like text and qnaid. I want to show the text in the button, and after clicking the button, the request will contain the qna id or more information in the prompt. Actually, i see the product in Test panel of QnA portal. I think that feature is what i exactly want. Could you please help?
image

PR #1408, opened 2019-04-16
[DotNet] Fixed StyleCop warnings in all projects

GasparAcevedoZainSouthworks:
## Proposed Changes
- Fix syntax that's non-compliant with StyleCop guidelines

- Set to ignore rule [SA0001](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA0001.md)  by modifying the ruleset files


GasparAcevedoZainSouthworks:
OK @johnataylor, we will close the PR.

PR #1413, opened 2019-04-17
Zerryth/dispatch sample with new pattern

Zerryth:
Creating parity between C# and Node samples by updating `14.nlp-with-dispatch` on the Node side.

_Note: creating this as a Draft PR as discussion needs to occur regarding the readmes for both C# & Node for this sample_

## Proposed Changes
  - created recognizers for luis and qna in `index.js`, then pass them into constructor of `DispatchBot` with `logger`
  - As done on the C# side, nlp-with-dispatch sample in node now can call child luis models from parent dispatch luis model, and results of children are collected from `RecognizerResult.connectedServiceResult`
  - bot now derives from new `ActivityHandler` class
  - event-based pattern (use of `onMessage()` and `onMembersAdded` in bot)
  - calls new `bot.run()` method in post calls to `/api/messages`


Zerryth:
Oops.

PR #1414, opened 2019-04-17
[Sample 14.nlp-with-dispatch] node update

Zerryth:
Creating parity between C# and Node samples by updating `14.nlp-with-dispatch` on the Node side.

_Note: creating this as a Draft PR as discussion needs to occur regarding the readmes for both C# & Node for this sample_

## Proposed Changes
  - created recognizers for luis and qna in `index.js`, then pass them into constructor of `DispatchBot` with `logger`
  - As done on the C# side, nlp-with-dispatch sample in node now can call child luis models from parent dispatch luis model, and results of children are collected from `RecognizerResult.connectedServiceResult`
  - bot now derives from new `ActivityHandler` class
  - event-driven pattern (use of `onMessage()` and `onMembersAdded` in bot)
  - calls new `bot.run()` method in post calls to `/api/messages`


Zerryth:
Bot code runs as expected--pending readme update
Also answer to discussion going on in email regarding the intent name related to qna service

samples/javascript_nodejs/14.nlp-with-dispatch/.env
```diff
@@ -1,2 +1,18 @@
 MicrosoftAppId=
 MicrosoftAppPassword=
+
+Luis-Home-Automation-AppId=
```
sgellock: We should be consistent in how we use keys here.  With some keys were are using PascalCase which is the pattern, and with other keys you're using kabob-case.  Please use PascalCase for Luis keys and QnA keys, and NlpWithDispath... keys

samples/javascript_nodejs/14.nlp-with-dispatch/bots/dispatchBot.js
```diff
@@ -0,0 +1,112 @@
...
+            this.logger.log('Processing Message Activity.');
+
+            // First, we use the dispatch model to determine which cognitive service (LUIS or QnA) to use.
+            const recognizerResult = await dispatchRecognizer.recognize(context);
+
+            // Top intent tell us which cognitive service to use.
+            const topIntent = LuisRecognizer.topIntent(recognizerResult);
+
+            // Next, we call the dispatcher with the top intent.
+            await this.dispatchToTopIntentAsync(context, topIntent, recognizerResult);
```
sgellock: don't you need to call await next(); as the last call of your onMessage function?

Zerryth: ...YES...

samples/javascript_nodejs/14.nlp-with-dispatch/.env
Zerryth: PascalCase it is! This original in the C# version made my soul cry: "`Luis-nlp-with-dispatchDispatch-Authoringkey`"

Not only that, but 1. js doesn't like "with" (reserved keyword) and 2. node/.env doesn't like `-` in `.env` files apparently

Removed `-`'s and switched to Pascal :thumbsup:

samples/javascript_nodejs/14.nlp-with-dispatch/.env
```diff
@@ -1,18 +1,18 @@
...
-Luis-Home-Automation-Authoringkey=
-Luis-Home-Automation-Region=
+LuisHomeAutomationAppId=
+LuisHomeAutomationAuthoringkey=
+LuisHomeAutomationRegion=
 
-sampleQnaKbId=
-sampleQnaEndpointKey=
-sampleQnaHostname=
+SampleQnaKbId=
```
stevengum: Can we change this to the following? 
QnAKnowledgebaseId,
QnAAuthKey,
QnAEndpointHostName

This mirrors the environment variable names that the QnA Maker team is using in their bot templates.

samples/javascript_nodejs/14.nlp-with-dispatch/.env
```diff
@@ -1,2 +1,18 @@
 MicrosoftAppId=
 MicrosoftAppPassword=
+
+LuisHomeAutomationAppId=
```
stevengum: Does the user still need this information if it's already hooked up to the Dispatch model?

samples/javascript_nodejs/14.nlp-with-dispatch/cognitiveModels/Weather.json
```diff
@@ -77,7 +77,7 @@
       ]
     },
     {
-      "text": "forecast in celcius",
+      "text": "forecast in celsius",
```
stevengum: Thanks for catching this!

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
```diff
@@ -8,46 +8,41 @@ const path = require('path');
 const restify = require('restify');
 
 // Import required bot services. See https://aka.ms/bot-services to learn more about the different parts of a bot.
-const { BotFrameworkAdapter, MemoryStorage, ConversationState, UserState } = require('botbuilder');
-// Import required bot configuration.
-const { BotConfiguration } = require('botframework-config');
+const { BotFrameworkAdapter } = require('botbuilder');
+const { LuisService } = require('botframework-config');
```
stevengum: Since we've removed the .bot file, we should also remove the dependency on `botframework-config`. I realize this differs from the C# sample, but the `LuisService` is just a data model. 

We don't have the `botframework-config` dependency in `botbuilder-ai` that `Microsoft.Bot.Builder.Ai.LUIS` has on `Microsoft.Bot.Configuration`, so let's get rid of the dependency.

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
```diff
@@ -8,46 +8,41 @@ const path = require('path');
...
+
+const luisService = new LuisService({
+    appId: process.env.DispatchAppId,
+    authoringKey: process.env.DispatchAuthoringKey,
+    region: process.env.DispatchRegion
+});
+
+const dispatchRecognizer = new LuisRecognizer({
+    applicationId: luisService.appId,
+    endpoint: luisService.getEndpoint(),
```
stevengum: Can you instead use the similar code/environment variable from Core-Bot?

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
```diff
@@ -8,46 +8,41 @@ const path = require('path');
...
+    endpointKey: luisService.authoringKey,
+}, {
+    includeAllIntents: true,
+    includeInstanceData: true
+}, true);
+
+const qnaMaker = new QnAMaker({
+    knowledgeBaseId: process.sampleQnaKbId,
```
stevengum: This should be `process.env.QnAKnowledgebaseId` etc, you're missing the `env`

samples/javascript_nodejs/14.nlp-with-dispatch/.env
Zerryth: No problem, changed away!

Although with `QnAAuthKey` specifically I'm personally not a huge fan of, because imagine it from a user's perspective, you're told to find the `QnAAuthKey`, there's nothing labeled with that specific label. So the user could think "Is this the key in Azure portal? Or maybe it's something in qnamaker.ai..."
It's specifically the EndpointKey in the settings page at qnamaker.ai and 

image


So imo `QnAEndpointKey` would be better...but I have it as `QnAAuthKey` to mimic the QnA team's bot template :)

samples/javascript_nodejs/14.nlp-with-dispatch/.env
Zerryth: Sooo...there's a thread in email discussing the inclusion/exclusion of the .json file that's of the parent luis model that dispatches to the children services that it's hooked up to (Kamran and someone else he looped in) 

Issue with it is that to map to the children, it needs to have app ids (which have been checked in mistakenly from C# sample)

But the idea originally that Kamran wanted was to include the .json of the children so that the user can walk through each individual step of creating the services (luis, qna) individually + dispatch for full explanation of the pieces.

But then also include the parent luis model that dispatches to quickly get the bot running...

Anyways! 
TL;DR TBD on what CognitiveModels will actually be included in samples

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
Zerryth: aye-aye!

removing

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
Zerryth: Ok. I'll change the variable names once the email convo has resolved, since it will determine whether or not the env need to be more specific (i.e. if there are still 3 luis env variables that need to look different from one another)

samples/javascript_nodejs/14.nlp-with-dispatch/index.js
```diff
@@ -8,46 +8,41 @@ const path = require('path');
...
-const DEV_ENVIRONMENT = 'development';
-
-// bot name as defined in .bot file or from runtime
-const BOT_CONFIGURATION = (process.env.NODE_ENV || DEV_ENVIRONMENT);
-
-// Get bot endpoint configuration by service name
-const endpointConfig = botConfig.findServiceByNameOrId(BOT_CONFIGURATION);
-
 // Create adapter.
 // See https://aka.ms/about-bot-adapter to learn more about .bot file its use and bot configuration .
```
sgellock: Can you remove the text about .bot files?  we no longer use or expect to talk to customers about .bot file usage and configuration.  I think the aka.ms link is fine to keep

Zerryth: Sneaky devil, be gone~
fixed, ty

samples/javascript_nodejs/14.nlp-with-dispatch/.env
Zerryth: It was decided that the users will simply follow the tutorial. This sample does not have a super fast plug-and-play option, due to the nature of the dispatch jsons needing to map to the children with appIds--which we won't check in

Deleted usused cognitive models & unusued .env variables

Zerryth:
@stevengum pushed changes you requested (changed variable names to match corebot, got rid of the children luis variables). Take a gander again!

PR #1426, opened 2019-04-22
Adding base path to service url to fix 404 error

jspruance:
Fixes #
This PR contains a fix for the issue described here:
https://github.com/Microsoft/BotBuilder-V3/issues/133

## Proposed Changes
I found that the 404 error was a result of the base path ('/amer/') being omitted from the service URL. 

This fix was to set the base path using the Swagger 'setBasePath' method:
client.setBasePath(serviceUrl.path);

## Testing
No new tests should be needed for this change - manual testing results in the expected behavior of the bot returning the conversation members.

### DIFF
--- a/Node/core-GetConversationMembers/app.js
+++ b/Node/core-GetConversationMembers/app.js
@@ -107,10 +107,13 @@ function addTokenToClient(connector, clientPromise) {
 function printMembersInChannel(conversationAddress, members) {
     if (!members || members.length === 0) return;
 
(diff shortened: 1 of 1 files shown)


jspruance:
I found another preexisting bug while regression testing - name undefined. Fixed this by setting a default val of empty string and refactoring name printing method (110-114).

jspruance:
Retested after fix merge and conversation members are now being listed as expected.

Issue #1428, opened 2019-04-22
Broken link to install CLI tools in dotnet welcome user readme


ericksoen:
## Sample information
1. Sample type: samples
2. Sample language: dotnetcore
3. Sample name: 03.welcome-user

## Describe the bug
Readme links takes you to 404 page not found.

## To Reproduce
In the section [Deploy this bot to Azure](https://github.com/Microsoft/BotBuilder-Samples/tree/master/samples/csharp_dotnetcore/03.welcome-user#deploy-this-bot-to-azure) click the link to **Install CLI Tools**. It takes you to a 404 not found page.

## Expected behavior
Link takes you to the instruction page for installing CLI tools

## Screenshots
N/A

## Additional context
N/A

[bug]

sgellock:
this is fixed in the samples WIP branch

PR #1429, opened 2019-04-22
update readme link to use same case as file path

ericksoen:
Fixes [Issue #1428](https://github.com/Microsoft/BotBuilder-Samples/issues/1428)

## Proposed Changes
Update the casing in the ReadMe to match the casing of the file path. Locally, the link navigation works locally regardless of case, but on remote, it returns 404 not found if the case does not work.

## Testing
N/A


sgellock:
@ericksoen thank you for the PR.  This issue has been addressed in a previous PR that was merged.

PR #1438, opened 2019-04-24
Removed conditional checking membersAdded to match other samples.

### LINKED ISSUE #1436: Work in progress welcomeBot sample double check members being added
In the [samples-work-in-progress welcomeBot](https://github.com/Microsoft/BotBuilder-Samples/blob/samples-work-in-progress/samples/javascript_nodejs/03.welcome-users/bots/welcomeBot.js#L32) sample, we seem to be double-checking if members have been added:

```js
        this.onMembersAdded(async (context, next) => {
            // Do we have any new members added to the conversation?
            if (context.activity.membersAdded.length !== 0) {
            ...
```
Am I wrong to believe that the callback passed in to `onMembersAdded` should only run when members have been added? If this is the case, shouldn't `context.activity.membersAdded.length` always be greater than 0? 

Please let me know if I'm misunderstanding the API. Also, I know these samples are WIP, so if there's another way you'd like feedback, let me know!

[bug]

tdurnford:
Fixes #1436 

## Proposed Changes

Removed the if statement in onMembersAdded  in the Welcome Users and Suggested Actions samples that checks if the length of `context.activity.membersAdded` is not equal to zero. The ActivityHandler will only execute the function if a member has been added, so `context.activity.membersAdded.length !== 0`  will always be truthy. 

### 03.welcome-users
Annotation 2019-04-24 113045

### 08.suggested-actions
Annotation 2019-04-24 113300

### DIFF
--- a/samples/javascript_nodejs/03.welcome-users/bots/welcomeBot.js
+++ b/samples/javascript_nodejs/03.welcome-users/bots/welcomeBot.js
@@ -65,7 +65,6 @@ class WelcomeBot extends ActivityHandler {
         // Sends welcome messages to conversation members when they join the conversation.
         // Messages are only sent to conversation members who aren't the bot.
         this.onMembersAdded(async (context, next) => {
-           
             // Iterate over all new members added to the conversation
             for (let idx in context.activity.membersAdded) {
                 // Greet anyone that was not the target (recipient) of this message.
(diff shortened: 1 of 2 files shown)


stevengum:
Thanks for the fix and the screencaps!

PR #1448, opened 2019-04-26
[23.facebook-events] catch serialization exception

EricDahlvang:
Catch JsonSerializationException (which can occur if using the emulator, or a channel other than facebook messenger) and provide feedback to the user: "This sample is intended to be used with a Facebook bot."

### DIFF
--- a/samples/csharp_dotnetcore/23.facebook-events/Bots/FacebookBot.cs
+++ b/samples/csharp_dotnetcore/23.facebook-events/Bots/FacebookBot.cs
@@ -9,6 +9,7 @@ using Microsoft.Bot.Builder.Dialogs.Choices;
 using Microsoft.Bot.Schema;
 using Microsoft.BotBuilderSamples.FacebookModel;
 using Microsoft.Extensions.Logging;
+using Newtonsoft.Json;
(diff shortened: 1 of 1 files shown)


samples/csharp_dotnetcore/23.facebook-events/Bots/FacebookBot.cs
```diff
@@ -61,40 +62,51 @@ private async Task<bool> ProcessFacebookPayload(ITurnContext turnContext, object
...
+                        await OnFacebookEcho(turnContext, facebookPayload.Message, cancellationToken);
+                        return true;
+                    }
+
+                    // TODO: Handle other events that you're interested in...
+                }
+            }
+            catch(JsonSerializationException e)
+            {
+                if (turnContext.Activity.ChannelId != Bot.Connector.Channels.Facebook)
```
gabog: I would just re-throw if this condition is not met in case FB changes the schema for real in the future and we still get a serialization exception.

EricDahlvang: Good idea!  addressed

PR #1470, opened 2019-05-02
Fix active learning bot default thresholds

rokulka:
Fix active learning bot default thresholds


johnataylor:
thanks for the fix we are doing this PR against the actual sample

PR #1473, opened 2019-05-02
samples and generator updates for 4.4 release

sgellock:
- update all samples, all languages, all platforms
- update VSIX and yeoman code generators
- add .NET Templates
- update master README.md
- update SPEC.md
- update INSTALLING_CLI_TOOLS.md


sgellock:
@edjez @homezcx, samples 12 didn't have much conceptual difference with what we show in sample 13, the CoreBot sample.  So to simply the number of samples and attempt to provide samples that are differentiated we removed it.  If we evolve CoreBot at some point where it doesn't have LUIS (which we don't plan to do at this time), then we'll have a specific sample focused on NLP with LUIS.  To find the specific sample you mention, you can find that here:  https://github.com/microsoft/BotBuilder-Samples/tree/1b93328a6f86f3d8e40712f438fed65c07605f80/samples/csharp_dotnetcore/12.nlp-with-luis.

edjez:
👍 thx we'll see if we can adapt to sample 13 instead. 
@homzcx fyi

PR #1501, opened 2019-05-08
Adding node v3 / v4 migration sample

jspruance:
This PR adds a v3 and v4 version of the 'core-multiDialogs' bot to be used as a reference for developers who need to migrate their v3 applications to v4. 

v3 version is from the sample found here:
https://github.com/microsoft/BotBuilder-Samples/tree/v3-sdk-samples/Node/core-MultiDialogs

v4 version contains the same logic as v3 but has been upgraded to 4.4.0.


jspruance:
Question - maybe it is not necessary to actually have the v3 code here. Should we just reference it instead?

PR #1503, opened 2019-05-10
Refactor hotels dialog so that it has it's own class definition / fix…

jspruance:
Refactor 'hotels dialog' to have it's own class definition, similarly to how the 'flights' dialog is set up. This should provide better consistency for how we're initiating dialogs in the sample code. Also found and fixed a bug in the 'length of stay' display dates.


EricDahlvang:
The package.json should be updated as well:

>     "name": "echobot",
>     "description": "Bot Builder v4 echo bot sample",

MigrationV3V4/Node/core-MultiDialogs-v4/dialogs/hotels.js
```diff
@@ -1,60 +1,85 @@
...
 } = require('../const');
 
-const hotelsDialog = [
-    async (stepContext) => {
+class HotelsDialog extends ComponentDialog {
+    constructor(id) {
+        super(id);
+
+        this.addDialog(new WaterfallDialog(HOTELS_WATERFALL_DIALOG, [
+            this.destinationPromptStep,
```
EricDahlvang: Should this line have .bind(this) also?

EricDahlvang:
Flights dialog should throw an exception, which will be caught by adapter.onTurnError (demonstrating Bot error handling, as the message states).

MigrationV3V4/Node/core-MultiDialogs-v4/dialogs/hotels.js
```diff
@@ -1,60 +1,85 @@
...
             CHECKIN_DATETIME_PROMPT, {
                 prompt: 'When do you want to check in?'
             }
         );
-    },
-    async (stepContext) => {
+    }
+
+    async checkinTimeSetStep (stepContext) {
         const checkinTime = stepContext.result[1].value;
```
EricDahlvang: This should be stepContext.result[0].value;    (stepContext.result[1] is undefined)

jspruance:
> The other V4 node samples do not have the following files: .deployment, deploy.cmd, iisnode.yml, publish.cmd and web.config. The PostDeployScripts also do not exist in the V4 node samples.
> 
> Also, missing .env file with:
> MicrosoftAppId=
> MicrosoftAppPassword=

We typically wouldn't want to check in the .env file don't you think?

jspruance:
> The other V4 node samples do not have the following files: .deployment, deploy.cmd, iisnode.yml, publish.cmd and web.config. The PostDeployScripts also do not exist in the V4 node samples.
> 
> Also, missing .env file with:
> MicrosoftAppId=
> MicrosoftAppPassword=

Those files are generated from Azure, when creating the bot there. Do you think we should remove then for this sample even though we are recommending that they create their v4 bot through Azure?

MigrationV3V4/Node/core-MultiDialogs-v4/dialogs/hotels.js
jspruance: strange - i retested this and both index 0 and 1 contain the correct value. i must have been seeing something differently before to have used '1'. i'll change it to '0' for now.

MigrationV3V4/Node/core-MultiDialogs-v4/dialogs/hotels.js
jspruance: Added back.

jspruance:
> Flights dialog should throw an exception, which will be caught by adapter.onTurnError (demonstrating Bot error handling, as the message states).

Updated the code the throw an exception from flights dialog - the 'onTurnError' handler is now catching it and outputting the error mssg.

EricDahlvang:
I was just comparing this V4 sample to the other V4 samples in the master branch.  Every one of them contain a .env file.  In fact, it is required to run the sample. 

I'm of the opinion that we should not include the Azure generated files, since they are not pertinent to anything we are demonstrating related to migration.  Otherwise, users doing a comparison of the V3 vs V4 multi-dialog will be wondering what these other files are (when they really have nothing to do with migration).

@scheyal and @JonathanFingold do you have any input here?

jspruance:
Standard practice is to not commit ‘.env’ files to source control since they typically contain secret keys and environment specific configuration. This sample does work fine without it.

Ok, I’ll remove the Azure files.

MigrationV3V4/Node/core-MultiDialogs-v4/const.js
```diff
@@ -9,5 +9,7 @@ module.exports = {
     FLIGHTS_DIALOG: 'flightsDialog',
     SUPPORT_DIALOG: 'supportDialog',
     USER_PROFILE_PROPERTY: 'userProfileProperty',
-    CONVERSATION_STATE_ACCESSOR: 'conversationStateAccessor'
+    CONVERSATION_STATE_ACCESSOR: 'conversationStateAccessor',
+    HOTELS_WATERFALL_DIALOG: 'hotelsWaterfallDialog',
```
JonathanFingold: I'm not sure why you'd make the two _waterfall dialog IDs_ public constants. Each will be local within their respective component dialog, and nothing outside the component dialog will be able to make relevant use of it.

jspruance: agree...removed consts

PR #1536, opened 2019-05-27
Fix node Multi Lingual Bot translatorKey.  Echo back what the user entered

EricDahlvang:
Fixes #1535

PR #1566, opened 2019-06-10
Fix key name from translationKey to correct translatorKey [17.multilingual-bot]

nt-7:
## Fix env key name from translationKey to translatorKey on env
- Correct : translatorKey
- Error : translationKey

## References
- Call Place
https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/javascript_nodejs/17.multilingual-bot/index.js#L55
- Related Issue
https://github.com/microsoft/BotBuilder-Samples/pull/1528


nt-7:
Already fixed (https://github.com/microsoft/BotBuilder-Samples/pull/1536/files)

PR #1567, opened 2019-06-10
Encourage best practice by always forwarding available cancellation token to awaitable calls

chivandikwa:
## Proposed Changes
Make samples more complete by always passing down available cancellation tokens to any awaitable calls.


sgellock:
@gabog @johnataylor merge conflicts not withstanding, do you want to pick this up?  this PR isn't complete, as .NET Templates, and VSIX Templates would need to be modified in addition to the files in this PR.

johnataylor:
let's do this next time around in 4.8 - because we have to redo generators.
opened https://github.com/microsoft/BotBuilder-Samples/issues/2148 to track this

PR #1568, opened 2019-06-10
Timex resolution

chivandikwa:
## Proposed Changes
Minor code improvement:
- Replace hard coded access with loop
- Make it easy to read through samples by created named groups with spacing

formatted

### DIFF
--- a/samples/csharp_dotnetcore/40.timex-resolution/Program.cs
+++ b/samples/csharp_dotnetcore/40.timex-resolution/Program.cs
@@ -1,24 +1,37 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Collections.Generic;
+
 namespace Microsoft.BotBuilderSamples
 {
     class Program
     {
         static void Main(string[] args)
         {
-            // Creating TIMEX expressions from natural language using the Recognizer package.
-            Ambiguity.DateAmbiguity();
-            Ambiguity.TimeAmbiguity();
-            Ambiguity.DateTimeAmbiguity();
-            Ranges.DateRange();
-            Ranges.TimeRange();
+            var samples = new List<Action>
+            {
+                // Creating TIMEX expressions from natural language using the Recognizer package.
+                Ambiguity.DateAmbiguity,
+                Ambiguity.TimeAmbiguity,
+                Ambiguity.DateTimeAmbiguity,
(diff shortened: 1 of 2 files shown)


samples/csharp_dotnetcore/40.timex-resolution/Resolution.cs
```diff
@@ -20,21 +20,12 @@ public static void Examples()
...
-
-            Console.WriteLine(resolution.Values[3].Timex);
-            Console.WriteLine(resolution.Values[3].Type);
-            Console.WriteLine(resolution.Values[3].Value);
+            foreach (var value in resolution.Values)
+            {
+                Console.WriteLine(value.Timex);
+                Console.WriteLine(value.Type);
+                Console.WriteLine(value.Value);
+            }
```
johnataylor: The reason the 4 resolutions were explicitly called out in a fixed way is because this is a fixed size. So if you want to avoid code duplication the most correct factoring would be a for loop counting to 4 and not a foreach which implies 0 to n.

This resolution will result in 4 cases because there is ambiguity in the time part. This is mimicking LUIS which resolves dates backwards and forwards from now. So with ambiguity in time you get 4 resolutions otherwise you get 2.

samples/csharp_dotnetcore/40.timex-resolution/Program.cs
```diff
@@ -1,24 +1,37 @@
...
+                Resolution.Examples,
+                Constraints.Examples
+            };
 
-            // Manipulating TIMEX expressions in code using the TIMEX Datatype package.
-            Parsing.Examples();
-            LanguageGeneration.Examples();
-            Resolution.Examples();
-            Constraints.Examples();
+            foreach (var sample in samples)
```
johnataylor: I don't see how this is better or in any way more readable! Just replacing a sequential set of functions with a list of functions that is then iterated over. This is just taking a perfectly natural and simple mechanism of code and replacing it with an artificial handcrafted mechanism. So I don't think we need this part at all.

samples/csharp_dotnetcore/40.timex-resolution/Resolution.cs
johnataylor: So change this something like:
```
for (var i=0; i<4; i++) { }
```
and it works. Otherwise I prefer the code that's there, it being more accurate.

Clearly this also needs more comments - I can add that.

PR #1570, opened 2019-06-12
4.5.1 Corebot updates

johnataylor:
It has been reported that people expect in Core bot the book details to contain the city if it matches the ML composite but misses the closed list.

This isn't exactly correct, the response from LUIS can be a little confusing when the composite matches but the entity doesn't; clearly the model is stating a _closed set of airports_, the intention is that a city not in this list is not somewhere you can fly to or from. The good news is that this case can be identified in the LUIS response. Specifically the child of the composite is empty but the entity is there. The correct behavior in the bot is therefore to spot this as an error and send that fact to the user.

The fact that later inline recognizers don't honor the closed nature of the list is really also a bug but we've left that as an exercise for the application developer. We could add a comment to include a PromptValidator; however the real underlying problem is that the model is split between code and the LUIS model.


samples/javascript_nodejs/13.core-bot/package.json
```diff
@@ -19,7 +19,6 @@
         "@microsoft/recognizers-text-data-types-timex-expression": "^1.1.4",
         "botbuilder": "^4.5.0-preview.69412",
```
gabog: @stevengum, do we need an explicit ref to botbuilder? or is it implicit in botbuilder-dialogs, ai and testing?

stevengum: Good ask; we do need the explicit ref to `botbuilder`. If this dependency is referenced in the other libraries it isn't exposed in the same way that `botframework-schema` and `botbuilder-core` is exposed through `botbuilder`.

gabog: k, thanks

PR #1572, opened 2019-06-12
Adding required versionId field, update bing geography entity

jspruance:
The sample LUIS app in the 'intelligence-LUIS' Node sample doesn't pass validation, so I am making the necessary updates:

1) Add a 'versionId' filed
2) Update 'bing_entities' value from 'geography' > 'geographyV2'


jspruance:
Confirmed - this change is needed in order to successfully import the sample LUIS app. Merging.

Issue #1577, opened 2019-06-15
Core bot (via ABS) home page between JS and C#


diberry:
When creating ABS Basic bot v4 C# (this is supposed to be the core bot #13), when running bot on local host computer, the root website returns a nice home page letting me know the bot website works. When I create ABS Basic bot v4 Node.js, and run it on local host computer, there isn't a nice home page letting me know the bot website works. Can one be added? Or is it there and just not working?

benbrown:
It is not there. One could be added!

diberry:
Created [PR](https://github.com/microsoft/BotBuilder-Samples/pull/1582) for this.

sgellock:
I'm moving this out to the MQ milestone

sgellock:
We've decided not to do this work as it adds route code that will take away form the intended purpose of the samples

PR #1582, opened 2019-06-17
Core bot node.js - return home page - like C# core bot does

diberry:
Fixes #

Issue #1577 1577

## Proposed Changes
Copied the home page from core c#.
Added route for / to return that file.

## Testing
I could add a test but I don't see any tests for this bot. Where are those?

### DIFF
--- a/samples/javascript_nodejs/13.core-bot/default.htm
+++ b/samples/javascript_nodejs/13.core-bot/default.htm
@@ -1,0 +1,417 @@
+﻿<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>Core Bot Sample</title>
+    <style>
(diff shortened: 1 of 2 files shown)


sgellock:
This change, if we decided to take it, needs a few things addressed.  One issue is there is a more simpler way in Node.js to send the http get response for the home page.  this code is using older patterns.  The second issue is that html page could use some cleanup.  there is some markup and css cleanup that it could benefit from.  Thirdly, this is only one sample.  if we were to seek parity with our C# sample web app behavior, then the other ~35 JavaScript, TypeScript, Node.js and Yeoman samples will need to include this.

johnataylor:
Teh point about consistency is well taken, however, we have decided adding a home page to the sample doesn't really add much and is tangential to what we are trying to show. We think node developers know how to serve up static pages - or can easily learn from some other source.
Besides which this change as it is is only a small part of the work that is required - because the generators would also need updating.

PR #1587, opened 2019-06-20
Update the usage of TemplateEngine

feich-ms:
Update the usage of TemplateEngine. Currently, template engine will use new TemplateEngine().AddFiles() to load lg files


feich-ms:
close due to dup

Issue #1609, opened 2019-06-28
[README.md] Duplicate about CLA


joconte:
Just a duplicate in README.md :

```
When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA.

When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repos using our CLA. This project has adopted the Microsoft Open Source Code of Conduct. For more information see the Code of Conduct FAQ or contact opencode@microsoft.com with any additional questions or comments.
```

PR #1610, opened 2019-06-28
Fix a duplicate about CLA in README.md

joconte:
Fixes https://github.com/microsoft/BotBuilder-Samples/issues/1609


sgellock:
@joconte thanks for the PR.

PR #1633, opened 2019-07-09
add credential provider back to dotnet proactive bot startup

EricDahlvang:
It was mistakenly removed with this commit: 
https://github.com/microsoft/BotBuilder-Samples/commit/31eb77e5923f48906844427a1b2a299bb1cadaf5#diff-633d7e7e1a7f5e883042736fbf9d039a

### DIFF
--- a/samples/csharp_dotnetcore/16.proactive-messages/Startup.cs
+++ b/samples/csharp_dotnetcore/16.proactive-messages/Startup.cs
@@ -21,6 +21,9 @@ namespace Microsoft.BotBuilderSamples
         {
             services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
 
+            // Create the credential provider to be used with the Bot Framework Adapter.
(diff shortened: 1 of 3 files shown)


samples/csharp_dotnetcore/16.proactive-messages/Startup.cs
```diff
@@ -21,6 +21,9 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
 
+            // Create the credential provider to be used in the NotifyController.
+            services.AddSingleton<ICredentialProvider, ConfigurationCredentialProvider>();
```
johnataylor: if this is using the overload of the constructor:

```
protected BotFrameworkHttpAdapter(IConfiguration configuration, ILogger<BotFrameworkHttpAdapter> logger = null)
```

Then we shouldn't need the explicit IConfigurationProvider on the services collection???

PR #1634, opened 2019-07-09
added inspection sample

johnataylor:
This is just the C# sample


johnataylor:
@cwhitten can you take a quick look at the README instructions. I was using two emulators (which worked great) but I think the steps can be simpler.

PR #1639, opened 2019-07-10
inspection sample in JavaScript

johnataylor:


johnataylor:
The README needs a little more work. It needs to have clear instructions as to how to use the latest side-car debugging mode in the emulator. I tested with two emulators and explicitly typed in the /INSPECT open commend.

PR #1657, opened 2019-07-17
Fix browser echo sample for IE11

### LINKED ISSUE #1618: javascript_es6/01.browser-echo sample throws an error in Edge
## Sample information
1. Sample language:  es6
2. Sample name: javascript_es6/01.browser-echo

## Describe the bug
The sample '01.browser-echo' works fine on Chrome but throws up an error on Edge.
**Error**: "SCRIPT1028: SCRIPT1028: Expected identifier, string or number
app.js (16292,24)"

The issue is with the use of object spread operator which is not supported in Edge : 
const card = { ...other };

## To Reproduce
1. Run the sample '01.browser-echo' on the Command Prompt.
2. Open Edge and type 'http://localhost:8080' and enter.
3. Page doesn't show anything and is blank.
4. Press F12 to inspect element and you will see the error on the console "SCRIPT1028: SCRIPT1028: Expected identifier, string or number
app.js (16292,24)".

## Additional context
- Tried using  @babel/plugin-transform-spread and then  including  @babel/plugin-transform-spread and @babel/plugin-proposal-object-rest-spread in the presets in .babelrc file, but it still shows the same error.
 - The targets for the presets in .babelrc file only specifies for the Chrome version but not for Edge.

[bug]

compulim:
Semi-fix #1618.

## Proposed Changes

- Update all dependencies to latest version
   - Bump to `botbuilder-core@4.5.0` and `botframework-webchat@4.5.0`
   - Removing unnecessary dependencies and plugins
- Configure Babel to target "minus 2 versions" and Internet Explorer 11
- Configure Webpack to transpile `botbuilder-core` because the package is [targeting ESNext instead of ES5](https://github.com/microsoft/botbuilder-js/blob/master/libraries/botbuilder-core/tsconfig.json#L3)
- Use `core-js` and `regenerator-runtime` instead of deprecated `@babel/polyfill`

In order to fix this sample in IE11, we requires 3 more fixes from `botbuilder-js` project:

- https://github.com/microsoft/botbuilder-js/issues/1073
- https://github.com/microsoft/botbuilder-js/issues/1075
- https://github.com/microsoft/botbuilder-js/issues/1076

#### `turnContext.js/removeMentionText`

```diff
- var mentionNameMatch = mentions[i].text.match(/(?<=<at.*>)(.*?)(?=<\/at>)/i);
+ var mentionNameMatch = [];
```

#### `internal.js`

```diff
- if (Proxy && Proxy.revocable) {
+ if (typeof Proxy !== 'undefined' && Proxy.revocable) {
```

## Testing

1. Test using Webpack development server
   1. `npm start`
   1. Browse to http://localhost:8080/ under IE11 and modern browsers
1. Test using Webpack production build
   1. `npm run build`
   1. `npx serve dist`
   1. Browse to http://localhost:5000/ under IE11 and modern browsers


samples/javascript_es6/01.browser-echo/package.json
```diff
@@ -1,47 +1,42 @@
...
         "lint": "eslint .",
         "test": "echo \"Error: no test specified\" && exit 1"
     },
     "dependencies": {
-        "@babel/runtime": "^7.1.5",
-        "botbuilder-core": "^4.2.0",
-        "botframework-directlinejs": "~0.11.2",
-        "botframework-webchat": "4.2.0",
-        "core-js": "^3.0.1"
+        "botbuilder-core": "^4.5.0",
```
stevengum: Can you use `~` instead of `^`? We test the versions by minor, so 4.2 might not necessarily work with 4.5.

PR #1658, opened 2019-07-17
Use standard tts voice

dargilco:
## Proposed Changes
Use standard voice instead of Neural voice, since Neural voices are only available in limited Azure regions. For example, developers who are using free-trial speech subscriptions will not be able to use this voice. In my tutorial I will address this and show how to use Neural. But the checked-in sample code should use Standard voice until we have more wide-spread deployment of Neural voices.

## Testing
Compiled, deployed bot. Tested with DL Speech Client tool with west-us speech keys to make sure the Bot can speak.


dargilco:
@Jeffders  @cleemullins please review. For some reason I don't see a way to add Reviewers on the right-side options.

PR #1676, opened 2019-07-30
Added documentation on how to set up in Teams

BobGerman:
Fixes confusion when using the sample in Teams  #

## Proposed Changes
This is only a documentation change.

The sample did not work in Teams until I figured out that the token login service domain (token.botframework.com) needs to be listed as valid in the Teams manifest. I hope to save others the time I spent figuring that out.

## Testing
n/a


johnataylor:
This is a Teams specific sample and should follow the pattern we've now established with the other Teams specific samples. It should also include a templated manifest.json as we have in the others.

johnataylor:
opened tracking issue https://github.com/microsoft/BotBuilder-Samples/issues/2028

clearab:
Hi @BobGerman - I've incorporated this into #2149, we wanted to align the manifest with the other Teams samples, while incorporating your inclusion of the Graph API client and emphasis on the importance of token.botframework.com in the manifest. Let me know if you think this doesn't fulfill

PR #1680, opened 2019-07-31
Added new Active learning sample

trojenguri:
Active learning sample on latest architecture and with latest SDK data.
Support for nodejs added
Remove it from experimental.
Update the documentation.


trojenguri:
> Parity with JS required.

NodeJS is also added

PR #1696, opened 2019-08-07
QnAMaker-App-Insights Sample App

WashingtonKayaker:
Sample app to show QnA Maker with Telemetry integrated


samples/csharp_dotnetcore/22.qnamaker-App_insights/Bots/QnABot.cs
```diff
@@ -0,0 +1,58 @@
...
+            _logger = logger;
+            _httpClientFactory = httpClientFactory;
+            _telemetryClient = telemetryClient;
+        }
+
+        protected override async Task OnMessageActivityAsync(ITurnContext<IMessageActivity> turnContext, CancellationToken cancellationToken)
+        {
+            var httpClient = _httpClientFactory.CreateClient();
+
+            var qnaMaker = new MyQnAMaker(new QnAMakerEndpoint
```
munozemilio: Can we inject this through DI as well?

WashingtonKayaker: Eyal in his email Sent on Monday, August 26, 2019 12:58 PM, Suggested leaving it as is.  If you do not agree, we should hash it out.

WashingtonKayaker:
@sgellock  

> At the very least the folder name for this sample needs to conform to the other folder names in the samples/csharp_dotnet folder.  the folder name should be all lower case, and use "-" as word separators.  We don't use "_" underscores or capital litters in the folder names.

I fixed the folder name.

johnataylor:
We don't think we want to take this into our samples set at this stage. We already feel we have quite a lot of samples. If you feel that you are illustrating some important areas that are not covered, please open an issue for the specific cases. Thanks.

PR #1755, opened 2019-09-10
custom recognizer sample for nodejs

jspruance:
Adding a sample with a custom recognizer to aid with V3-V4 migrations.

### DIFF
--- a/MigrationV3V4/Node/custom-recognizer-example/bot.js
+++ b/MigrationV3V4/Node/custom-recognizer-example/bot.js
@@ -16,20 +16,21 @@ class MyBot extends ActivityHandler {
             const intent = recognizerResult.intents.length > 0 ? recognizer.getTopIntent(recognizerResult) : '';
             
             // respond according to the top intent provided by the recognizer
-            switch(intent) {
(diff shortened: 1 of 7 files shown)


MigrationV3V4/Node/custom-recognizer-example/bot.js
```diff
@@ -0,0 +1,52 @@
...
+                await context.sendActivity('Your response includes an email address. Would you like to contact us by email?');
+                break;
+              case 'SearchForEmailAddress':
+                await context.sendActivity('Your response includes an email address. Would you like to search for an email address?');
+                break;
+              case 'OpenEmailAccount':
+                await context.sendActivity('Your response includes an email address. Would you like to open an email account?');
+                break;
+              default:
+                await context.sendActivity('Your response includes an email address.');
```
EricDahlvang: Since this is the default, won't it fire even if there is no email address in the response?

jspruance: Updated with a condition to only mention email if an intent was detected (ie the user input a valid email address).

PR #1756, opened 2019-09-10
Add provisioning script to echo bot

weswigham:
## Proposed Changes
This PR adds a provisioning script to echo bot, this way the bot can be provisioned on azure by following guided prompts with a simple `npm run provision` (without the full azure cli needing to be installed and pre-setup).

I looked into this because while the bot framework startup for local js development was pretty OK, the instructions for provisioning and deployment on azure were _awful_ (pages of discourse on the proper way to set things up most of which by and large did not matter a whit, especially in a non-enterprise setting) - one button (or one command, in this case) deploys for the common usecases are usually pretty desirable, so I set to this. Hopefully at least the idea that a guided local provisioning process that could simplify initial setup is useful to someone. If so, ideally all the samples should have a guided provisioning script like this (IMO).

script-capture

Problems with this implementation:
* The script needs to auth _3 separate times_ to collect the appropriate information from the user. This seems like a failure in azure's auth design to me; but maybe I just don't get how to request a token that'll work for all the domains I need on the first try (help would be appreciated here).
* A paired `npm run deploy` would be nice, too, especially for CI/CD scenarios. I figure I can invest in that after I've gotten some amount of buyin on the idea that maybe 4 pages of reading do not need to be a prerequisite to deploying a bot to the cloud.


sgellock:
@weswigham thanks for the PR.  I'm going to respectfully decline taking this.  there are a number of reasons.  Echo bot is one of over ~35 node.js samples, if you include typescript and add the yeomen generator as well.  but more importantly, we have built our documentation and our flows around using ARM, so this change picks up a number of Az dependencies I'd rather not hoist into my samples.

weswigham:
Are you declining it because you'd only take a change integrated into every sample (which I said would be an option if this seemed like a good idea), or because the dependencies? If I also added a `deploy` task, this'd have _less_ dependencies, as you'd _not require the azure cli be installed_ to provision and deploy, just the bits of the azure js sdk needed to deploy aad apps and resource templates (which is much smaller). Plus you'd have a deploy command usable with azure pipelines - right now a bot can really only be easily CI/CD'd using the legacy kudu continuous deployment, as I've yet to find even an azure cli command that does a full kudu deployment on its own.

Hopefully you'll at least keep in mind the feedback that the process for provisioning a new bot is still cumbersome, at least. I was very much hoping this would be the starting point for a discussion, rather than just being accepted/rejected. Naturally, I don't know your team's process, but being on GitHub I'd hoped such a thing could occur. The point of this is to be able to simplify those _pages of documentation_ of deployment to just `npm run provision` (and ideally `npm run deploy`). I would hope the BotFramework team would understand that anyone who wanted to get into writing a simple bot probably doesn't really want to get into the weeds on how to create a new AAD App on azure (and clearly to some degree that's understood, otherwise you'd not have an ARM template for the compute assets), and the weight of "but existing docs say something else", doesn't preclude simplifying those docs alongside the change.

sgellock:
@weswigham let's reopen this and we can discuss.  from your reply maybe the hast of my working through issues to land the 4.6 release made things more abrupt than necessary.  

I am going to label it for 4.7, because there's no way we'll land this in time for 4.6, as I'm in the final days of getting stuff wrapped up.

weswigham:
Is there any sort of thing I need to do, or am I just waiting on some kind of decision from y'all, then?

sgellock:
for now, let's see if we get additionally community feedback.  I won't pay any more attention to this though until I've got 4.6 out the door.  so early November is the next time I could fully engage.

christopheranderson:
Closing in favor having this kind of functionality in some common tool (as opposed to copy+pasted to a bunch of samples). Tracking issue here: https://github.com/microsoft/botframework-cli/issues/286

PR #1775, opened 2019-09-13
Simplified adaptive card task module

EricDahlvang:


EricDahlvang:
moving to botbuilder-dotnet

PR #1784, opened 2019-09-29
Log exceptions properly to get the stack trace

mvelosop:
This is not really a bug fix or a new feature, just a minor enhancement.

## Proposed Changes

Currently, when exceptions are catched in the `AdapterWithErrorHandler` only the message is logged  but having the stack trace is usually useful to solve the problem.

So this change is to just use the logger overload that logs the stack trace.

## Testing

No testing really needed here, it's just enough to check that it compiles.


cleemullins:
@sgellock Scott, you've been in this code this week. Is this a PR we can take?

sgellock:
@mvelosop thanks for the PR!

Issue #1719, opened 2019-08-21
State management samples need to show how to use Cosmos DB


sgellock:
## Sample information
1. Sample type: samples
2. Sample language: [dotnetcore, nodejs]
3. Sample name: 45.state-management

## Describe the bug
CSS would like to have a sample that shows how to use Cosmos Db and BlobStorage.  Our current state-management sample uses memoryStorage()

mdrichardson:
I'd recommend waiting until [this is fixed](https://github.com/microsoft/botframework-sdk/issues/5467), then we can also include Partition Keys and some general guidance around them. I've also got [a draft blog post](https://microsoft-my.sharepoint.com/:w:/p/v-micric/EYhDBMJxSItGldr_j9x_K_8B68fhEkszhL9-sJ1vCDwFXQ?e=Lw6j2L) ready-ish to launch for this as well.

cleemullins:
@garypretty The ball is in your court. We can push this to 4.7, if the referenced CosmosDB issue needs to be fixed first.

garypretty:
@cleemullins We have the new storage provider in now. I'll update the sample to show blob and the new cosmos usage and ready to go for 4.6.

PR #1789, opened 2019-10-02
Update state sample to show usage of Blob / CosmosDB storage.

garypretty:
@cleemullins @sgellock This change is in response to #1719. Can you comment as to if this meets your requirements for showing Cosmos / Blob usage.

I took the decision to have them all in the single state sample, with MemoryStorage still the default - so that the bot still runs OOTB with no changes, but then have commented implementations for Cosmos / blob.
I also updated the readme with more details for the user to understand that other storage layers are available (and recommended for production) and some notes on implementation.

### DIFF
--- a/samples/csharp_dotnetcore/45.state-management/README.md
+++ b/samples/csharp_dotnetcore/45.state-management/README.md
@@ -4,6 +4,41 @@ This sample demonstrates how to save user and conversation data in an ASP.Net Co
 The bot maintains conversation state to track and direct the conversation and ask the user questions.
 The bot maintains user state to track the user's answers.
 
+By default this bot uses MemoryStorage for Conversation and User state.
+Memory Storage is great for testing purposes, but in a production scenario you will need to use persistent storage.
+Startup.cs contains commented examples of using Azure Blob Storage or Cosmos DB instead.
+
+# Azure Blob Storage
+
+To use Azure Blob Storage, create a Blob Storage account in your Azure subscription. You can then use the following code to
+create your storage layer, passing in your Blob Storage connection string and a blob container name.  
+
+***Note: You do not need to create the container manually, the bot will create the container for you if it does not exist.***
+
+```cs
+    var storage = new AzureBlobStorage("<blob-storage-connection-string>", "bot-state");
+```
+
+# CosmosDB Storage
+
+To use CosmosDB Storage, you need to create a CosmosDB instance in your Azure subscription. You can then use the following code to
+create your storage layer.  
+
+***Note: It is your responsibility to create an appropriate database within your CosmosDB instance. However, you should **not**
+create the container within the database, as the bot will do this for you and ensure the container is configured correctly.***
+
+```cs
+    var cosmosDbStorageOptions = new CosmosDbPartitionedStorageOptions()
+    {
+        CosmosDbEndpoint = "<endpoint-for-your-cosmosdb-instance>",
+        AuthKey = "<your-cosmosdb-auth-key>",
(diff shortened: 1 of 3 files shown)


samples/csharp_dotnetcore/45.state-management/StateMangementBot.csproj
```diff
@@ -7,7 +7,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.Bot.Builder.Integration.AspNet.Core" Version="4.5.1" />
+    <PackageReference Include="Microsoft.Bot.Builder.Azure" Version="4.6.0-preview-191001-1" />
+    <PackageReference Include="Microsoft.Bot.Builder.Integration.AspNet.Core" Version="4.6.0-preview-191001-1" />
```
mdrichardson: I'm still C#-dumb, but should the sample use the preview packages? It would necessitate users add the MyGet feed, wouldn't it? If not, it's fine.

I know we don't have `CosmosDbPartitionedStorage` in the current NuGet packages and the bot won't run without them...this is more of a pre-merge comment.

garypretty: @mdrichardson You are correct. However, I am merging all of my 4.6 samples, once approved, into a single branch that tracks the 4.6 MyGet preview packages. Once the public packages are available, I will update the one branch and PR to master.

samples/csharp_dotnetcore/45.state-management/README.md
```diff
@@ -4,6 +4,41 @@ This sample demonstrates how to save user and conversation data in an ASP.Net Co
...
+***Note: It is your responsibility to create an appropriate database within your CosmosDB instance. However, you should **not**
+create the container within the database, as the bot will do this for you and ensure the container is configured correctly.***
+
+```cs
+    var cosmosDbStorageOptions = new CosmosDbPartitionedStorageOptions()
+    {
+        CosmosDbEndpoint = "<endpoint-for-your-cosmosdb-instance>",
+        AuthKey = "<your-cosmosdb-auth-key>",
+        DatabaseId = "<your-database-id>",
+        ContainerId = "<cosmosdb-container-id>"
```
mdrichardson: Not saying it should or needs to be this way, but what are your thoughts on hard-coding the strings like this vs. putting them in `appsettings.json` and then bringing them in through either [`ConfigurationBuilder`](https://github.com/microsoft/BotBuilder-Samples/blob/4f978577e358d193c4a63ffdab4d10c2fc4b4874/experimental/multilingual-luis/csharp_dotnetcore/Startup.cs#L33) and/or [`BotServices`](https://github.com/microsoft/BotBuilder-Samples/blob/b49159c1cf3b6bd57cf7edf6d8f70a6d72d5a840/samples/csharp_dotnetcore/14.nlp-with-dispatch/BotServices.cs)?

garypretty: I deliberately didn't use the app settings to keep things as obvious and focused as possible.  I would be interested in getting input from @cleemullins here to get thoughts though.

mdrichardson: I thought that might be the case and lean towards that direction. On the support side, I have trouble pointing people to the appropriate sample (which I'd prefer to "here's *my* code...") when they want to use appsettings for this kind of thing.

cleemullins: Using a code comment like this is in line w/ our existing samples and seems the correct approach.

PR #1818, opened 2019-10-17
Chrande/console echo/remove recursion

christopheranderson:
## Proposed Changes

The 01.ConsoleEcho samples for JS and TS used unnecessary recursion, and had other oddities about them.

  - Remove recursion in TS and JS samples
  - Cleanup JS import/export
  - Cleanup JS type hints

## Testing

No behavioral changes


christopheranderson:
Yeah, I had to do something kinda weird there for the JSDoc type hint to resolve nicely. I think it's fine as is, but I could try to find a different way to resolve that interface type.

PR #1859, opened 2019-10-23
minor type fix

siduppal:
Fixes #
Minor issue: echos "ADDGroupID" instead of "AADGroupID"
## Proposed Changes

  - Super simple string change
  -
  -


EricDahlvang:
Closing since these samples are no longer in the Experimental directory, but are in the BotBuilder-dotnet and BotBuilder-js projects.

I've duplicated this patch here: https://github.com/microsoft/botbuilder-dotnet/pull/2896

PR #1863, opened 2019-10-25
DO NOT MERGE: Enable WebSockets and HttpGet on DotNet Samples

DDEfromOR:
These changes add the lines app.useWebSockets() to startup.cs and HttpGet to BotController.cs for all of the DotNetCore samples, in order to have them support the Streaming extensions added in BotBuilder 4.6, which allows them to work with the DirectLine Speech channel without further code changes.

For the lone WebAPI sample WebSocket support is included by default, so the controller method only needed route decorators. (It previous had none but was named Post, allowing the auto-route to be created for Post, but not Get.)

### DIFF
--- a/samples/csharp_dotnetcore/02.echo-bot/Startup.cs
+++ b/samples/csharp_dotnetcore/02.echo-bot/Startup.cs
@@ -48,7 +48,7 @@ namespace Microsoft.BotBuilderSamples
 
             app.UseDefaultFiles();
             app.UseStaticFiles();
-
+            app.UseWebSockets();
             //app.UseHttpsRedirection();
             app.UseMvc();
         }
--- a/samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
+++ b/samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
@@ -60,7 +60,7 @@ namespace Microsoft.BotBuilderSamples
 
         [HttpPost]
         [HttpGet]
-        public async Task<HttpResponseMessage> Post()
+        public async Task<HttpResponseMessage> PostAsync()
         {
             var response = new HttpResponseMessage();
 
--- a/samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
+++ b/samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
@@ -58,6 +58,8 @@ namespace Microsoft.BotBuilderSamples
             _dialog = new MainDialog(bookingRecognizer, bookingDialog, _loggerFactory.CreateLogger<MainDialog>());
         }
 
+        [HttpPost]
+        [HttpGet]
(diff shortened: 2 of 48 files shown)


samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
```diff
@@ -58,6 +58,8 @@ static BotController()
             _dialog = new MainDialog(bookingRecognizer, bookingDialog, _loggerFactory.CreateLogger<MainDialog>());
         }
 
+        [HttpPost]
+        [HttpGet]
```
christopheranderson: Why do two separate attributes here?

samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
```diff
@@ -58,6 +58,8 @@ static BotController()
...
         public async Task<HttpResponseMessage> Post()
```
christopheranderson: As long as we're touching this file, we should consider adding the `Async` suffix here.

samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
DDEfromOR: The styleguide insisted, it apparently has different rules for netCore vs netFx.

samples/csharp_webapi/13.core-bot/Controllers/BotController.cs
DDEfromOR: Sounds good, I'm betting it was left off before in order to take advantage of the auto-verb-matching, but now that we're explicitly adding verb tags it shouldn't be needed.

johnataylor:
We don't want Web Sockets on by default.

Alternatives include leaving these lines in but commented them out. But this would have to be accompanied by an appropriate comment and links to all the appropriate documentation for Direct Line Speech.

Enabling Web Sockets will mean the bot would stop working for Teams.

samples/csharp_dotnetcore/02.echo-bot/Startup.cs
```diff
@@ -48,7 +48,7 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 
             app.UseDefaultFiles();
             app.UseStaticFiles();
-
+            app.UseWebSockets();
```
johnataylor: we do not want web sockets enabled by default.

DDEfromOR: Not enabling WebSockets will mean the bot would stop working for DirectLine Speech. If a bot is unable to support both channels we have a big problem.

DDEfromOR:
@johnataylor can you explain the conflict between Teams and WebSockets?

PR #1864, opened 2019-10-25
Hostname check updated.

trojenguri:
Host name validation check
Code fix lint


samples/csharp_dotnetcore/11.qnamaker/BotServices.cs
```diff
@@ -14,10 +14,25 @@ public BotServices(IConfiguration configuration)
...
                 KnowledgeBaseId = configuration["QnAKnowledgebaseId"],
                 EndpointKey = configuration["QnAEndpointKey"],
-                Host = configuration["QnAEndpointHostName"]
+                Host = GetHostname(configuration["QnAEndpointHostName"])
             });
         }
 
         public QnAMaker QnAMakerService { get; private set; }
+
+        private static string GetHostname(string hostname)
```
cleemullins: Consider a more robust mechanism for host name validation. The URI class seems a better choice.

Issue #1873, opened 2019-10-28
[DirectLine Speech] Update 4.6 samples to work with DirectLine Speech.


DDEfromOR:
At least the CoreBot and EchoBot samples for 4.6 need to be updated to work with DirectLine Speech by default. This involves touching 2 lines of code, but these changes must be coordinated with the SDK team to make sure they are included in associated updates to generators and templates.

Specifically:
In startup.cs modify Configure to include:
```
            app.UseDefaultFiles();
            app.UseStaticFiles();
            app.UseWebSockets();
            app.UseMvc();
```
In BotController.cs modify the route decorator above PostAsync to read:
```
        [HttpPost, HttpGet]
```

The following assets need to be updated:
- [x] VSIX
- [x] .NET Templates
- [x] DotNet Core EchoBot & CoreBot Samples

DDEfromOR:
Corebot and Echobot are updated in the linked PR. If there are conflicts with other channels or if we decide to update more than the two samples please either contact me or make updates ASAP.

sgellock:
These changes have been merged into the 4.6-Staging branch.

PR #1874, opened 2019-10-28
Enable DLS by default on echo and core samples.

DDEfromOR:
Fixes #https://github.com/microsoft/BotBuilder-Samples/issues/1873


DDEfromOR:
Based on wrong branch.

PR #1891, opened 2019-10-31
[QnA Maker] Consuming BotBuilderSamples.Utils CardBuilder class for d…

trojenguri:
Consuming BotBuilderSamples.Utils CardBuilder class for displaying prompts.
Prompts null check.


johnataylor:
This change is already in 4.6-Staging

PR #1899, opened 2019-11-01
Update lg reference

boydc2014:
Fixes lg reference broken.  
 code. -->


boydc2014:
> Can you also remove the nuget.config since these are all in nuget now?

done

PR #1902, opened 2019-11-01
[Teams] Link unfurling sample updates (dotnet)

EricDahlvang:
Remove messaging (since this is a Messaging Extension that does not send messages)

Remove Bot from manifest

Add deployment templates

Update readme

### DIFF
--- a/samples/csharp_dotnetcore/55.teams-link-unfurling/AdapterWithErrorHandler.cs
+++ b/samples/csharp_dotnetcore/55.teams-link-unfurling/AdapterWithErrorHandler.cs
@@ -18,9 +18,10 @@ namespace Microsoft.BotBuilderSamples
                 // Log any leaked exception from the application.
                 logger.LogError(exception, $"[OnTurnError] unhandled error : {exception.Message}");
 
-                // Send a message to the user
-                await turnContext.SendActivityAsync("The bot encounted an error or bug.");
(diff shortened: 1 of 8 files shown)


samples/csharp_dotnetcore/55.teams-link-unfurling/AdapterWithErrorHandler.cs
```diff
@@ -18,9 +18,10 @@ public AdapterWithErrorHandler(IConfiguration configuration, ILogger<BotFramewor
                 // Log any leaked exception from the application.
                 logger.LogError(exception, $"[OnTurnError] unhandled error : {exception.Message}");
 
-                // Send a message to the user
-                await turnContext.SendActivityAsync("The bot encounted an error or bug.");
-                await turnContext.SendActivityAsync("To continue to run this bot, please fix the bot source code.");
+                // Note: Since this Messaging Extension does not have the messageTeamMembers permission
```
Virtual-Josh: This is an open question. Do we want to cut out the commented out lines of this sample? I'm not opposed to having them, but it looks to be inconsistent with other samples. 

The TeamsMessagingExtensionSearch has the lines in, but doesn't have the messageTeamMembers permission in the manifest. 

The TeamsMessagingExtensionAction has the lines commented out, but also does not have the permission in the manifest.

PR #1979, opened 2019-11-14
[Streaming, Experimental] fix and refactor experimental DLS Node.js echo-bot

stevengum:
Fixes microsoft/botbuilder-js#1408

### DIFF
--- a/experimental/directline-speech/javascript_nodejs/02.echo-bot/index.js
+++ b/experimental/directline-speech/javascript_nodejs/02.echo-bot/index.js
@@ -27,8 +27,7 @@ server.listen(process.env.port || process.env.PORT || 3978, () => {
 // The three values below are required for use with Direct Line Speech.
 const adapterSettings = {
     appId: process.env.MicrosoftAppId,
-    appPassword: process.env.MicrosoftAppPassword,
-    enableWebSockets: true
+    appPassword: process.env.MicrosoftAppPassword
 };
 
 // Create adapter.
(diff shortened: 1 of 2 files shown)


experimental/directline-speech/javascript_nodejs/02.echo-bot/index.js
```diff
@@ -23,16 +23,20 @@ server.listen(process.env.port || process.env.PORT || 3978, () => {
     console.log('\nTo talk to your bot, open the emulator select "Open Bot"');
 });
 
-// Create adapter.
-// See https://aka.ms/about-bot-adapter to learn more about how bots work.
-const adapter = new BotFrameworkAdapter({
+// BotFrameworkAdapterSettings configuration object with enableWebSockets set to true.
+// The three values below are required for use with Direct Line Speech.
+const adapterSettings = {
```
DDEfromOR: Since we're creating two adapters and only one will be used with WebSockets, we could create two sets of settings and only set enableWebSockets on the adapter instantiated inside the scope of the GET request below.

stevengum: Good idea, I could make a new object in the GET `/api/messages` route and point to the previous settings for the AppId and AppPassword. 

```ts
const streamingAdapter = new BotFrameworkAdapter({
    ...adapterSettings,
    enableWebSockets: true
});
```

stevengum:
I've also updated the `EchoBot` class to have parity with the C# EchoBot changes in https://github.com/microsoft/BotBuilder-Samples/pull/1982

PR #1986, opened 2019-11-15
Added Node and C# AdaptiveCardPrompt Samples

mdrichardson:
Reference: https://github.com/microsoft/botframework-sdk/issues/5396

# Do not review `AdaptiveCardPrompt.ts/cs`. It will change. Please only review "bot code" for now, if anything.

...I'll have a separate PR for those and merge the finalized code into the sample (or it might be in the SDK by then)

[Additional Details](https://gist.github.com/mdrichardson/320e4e20aac6252c1acac602032c34a2)

## Proposed Changes

Adds C# and Node Samples for using the un-merged [AdaptiveCardPrompt](https://github.com/mdrichardson/botbuilder-js/blob/adaptiveCardPrompt/libraries/botbuilder-dialogs/src/prompts/adaptiveCardPrompt.ts)

I numbered these Sample "09" because it seemed to make the most sense, but I'm very open to changing that, as well as the name.

These *mostly* follow the flow of current samples, but I originally created them in May, so I may have missed some design changes.

At some point, I may update this to show all of the functionality, but this is a good start while that gets fleshed out.

## Notes

Both of these have class files for `AdaptiveCardPrompt` that are directly part of the sample (not in the SDK), since this is meant to be merged before we merge in `AdaptiveCardPrompt`

C# AdaptiveCardPrompt relies on a couple of pieces of the Dialogs package that are internal. It's modified a little in the sample so that the sample can stand on its own.

Node AdaptiveCardPrompt is a compiled version of `adaptiveCardPrompt.ts` and then edited a little for readability.

## Testing

image


mdrichardson:
Closing now that AdaptiveCardPrompt is in the Community Repos

PR #1998, opened 2019-11-20
Add attachment prompt to MultiTurnPromptBot .NET sample

### LINKED ISSUE #1799: Create a sample to show taking input attachments on a waterfall dialog
currently as i see , the handling attachment bot sample is more of a bot sample jumping different functions , it would be a better implementation if the same can be implemented in a waterfall-dialog setting .
attachments are a key part of the chat bot experience , the one i am developing for my company's project requires this . 

currently i am trying to find a way to switch bots with a waterfall , 
any guidance will help . thank you .

EDIT : Multi attachment handling isnt very clear in the handling attachment sample .

v-kydela:
Fixes #1799

## Proposed Changes

image

### DIFF
--- a/samples/javascript_nodejs/05.multi-turn-prompt/dialogs/userProfileDialog.js
+++ b/samples/javascript_nodejs/05.multi-turn-prompt/dialogs/userProfileDialog.js
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+const { MessageFactory } = require('botbuilder');
 const {
+    AttachmentPrompt,
     ChoiceFactory,
     ChoicePrompt,
     ComponentDialog,
@@ -12,8 +14,10 @@ const {
     TextPrompt,
     WaterfallDialog
 } = require('botbuilder-dialogs');
+const { channels } = require('botbuilder-dialogs/lib/choices/channel');
 const { UserProfile } = require('../userProfile');
 
+const ATTACHMENT_PROMPT = 'ATTACHMENT_PROMPT';
 const CHOICE_PROMPT = 'CHOICE_PROMPT';
 const CONFIRM_PROMPT = 'CONFIRM_PROMPT';
 const NAME_PROMPT = 'NAME_PROMPT';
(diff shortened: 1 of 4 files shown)


samples/javascript_nodejs/05.multi-turn-prompt/dialogs/userProfileDialog.js
```diff
@@ -98,15 +104,32 @@ class UserProfileDialog extends ComponentDialog {
...
 
         const msg = step.values.age === -1 ? 'No age given.' : `I have your age as ${ step.values.age }.`;
 
         // We can send messages to the user at any point in the WaterfallStep.
         await step.context.sendActivity(msg);
 
-        // WaterfallStep always finishes with the end of the Waterfall or with another dialog, here it is a Prompt Dialog.
+        if (step.context.activity.channelId === channels.msteams) {
+            // This attachment prompt example is not designed to work for Teams attachments, so skip it in this case
+            return await step.next(undefined);
```
mdrichardson: Might be beneficial to directly notify the user that this won't work in Teams

```suggestion
            await step.context.sendActivity('This sample does not work in the Teams channel. Please test this in a different channel');
            return await step.next(undefined);
```

samples/javascript_nodejs/05.multi-turn-prompt/dialogs/userProfileDialog.js
```diff
@@ -118,25 +141,58 @@ class UserProfileDialog extends ComponentDialog {
...
+            attachments.forEach(attachment => {
+                if (attachment.contentType === 'image/jpeg' || attachment.contentType === 'image/png') {
+                    validImages.push(attachment);
+                }
+            });
+
+            promptContext.recognized.value = validImages;
+
+            // If none of the attachments are valid images, the retry prompt should be sent.
+            return validImages.length;
```
mdrichardson:
```suggestion
            return !!validImages.length;
```

samples/javascript_nodejs/05.multi-turn-prompt/dialogs/userProfileDialog.js
v-kydela: @mdrichardson - The sample works fine in Teams, though. It just won't have the attachment prompt in it.

mdrichardson: @v-kydela Right. Maybe a different message, then. I think a common way of testing samples is to download it and run it without reading the code. If somebody happens to do that in Teams, they might be confused. They might benefit from a notification for it. Consider this a nit-picky comment

PR #2022, opened 2019-12-05
Fix [48] QnA Maker Active Learning Sample (dotnet core sample - QnA Maker)

nt-7:
Fix System.ArgumentExcepion on QnA Maker Active Learning Sample (C#)
(https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/48.qnamaker-active-learning-bot)

## Change Proposal
The first argument of stepContext.NextAsync is expected to be object, but it is cancellationToken.
So, the sample bot got System.ArgumentExcepion : 'result cannot be a cancellation Token' in DialogHelper.cs.

- Before : return await stepContext.NextAsync(cancellationToken);
- **After** : return await stepContext.NextAsync(**stepContext.Result,** cancellationToken);

## Exception
System.ArgumentExcepion : 'result cannot be a cancellation Token' (DialogHelper.cs)

## Reference
https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/csharp_dotnetcore/48.qnamaker-active-learning-bot/Dialog/DialogHelper.cs#L78

### DIFF
--- a/samples/csharp_dotnetcore/48.qnamaker-active-learning-bot/Dialog/DialogHelper.cs
+++ b/samples/csharp_dotnetcore/48.qnamaker-active-learning-bot/Dialog/DialogHelper.cs
@@ -75,7 +75,7 @@ namespace Microsoft.BotBuilderSamples
 
             stepContext.Values[QnAData] = new List<QueryResult>(filteredResponse);
             stepContext.Values[CurrentQuery] = stepContext.Context.Activity.Text;
-            return await stepContext.NextAsync(cancellationToken);
+            return await stepContext.NextAsync(stepContext.Context, cancellationToken);
         }
 
         private async Task<DialogTurnResult> FilterLowVariationScoreList(WaterfallStepContext stepContext, CancellationToken cancellationToken)


samples/csharp_dotnetcore/48.qnamaker-active-learning-bot/Dialog/DialogHelper.cs
```diff
@@ -75,7 +75,7 @@ private async Task<DialogTurnResult> CallGenerateAnswer(WaterfallStepContext ste
 
             stepContext.Values[QnAData] = new List<QueryResult>(filteredResponse);
             stepContext.Values[CurrentQuery] = stepContext.Context.Activity.Text;
-            return await stepContext.NextAsync(cancellationToken);
+            return await stepContext.NextAsync(stepContext.Context, cancellationToken);
```
johnataylor: OK definitely a bug to pass the cancellationToken in as the result. However, this is now passing the current TurnContext in as the result. That doesn't seem correct either.

The idea of the result is that it is the result of this step that arrives as the input value of the next step.

johnataylor:
this was fixed by another PR - thanks for sending this - sorry for the delay

PR #2043, opened 2019-12-13
add notes about TrustServiceUrl

mdrichardson:
## Proposed Changes

Adds additional notes to the README's and Proactive controllers about using `TrustServiceUrl()` to avoid permissions issue when sending proactive messages.

We [have documentation for this](https://docs.microsoft.com/en-us/azure/bot-service/bot-builder-howto-proactive-message?view=azure-bot-service-4.0&tabs=csharp#avoiding-401-unauthorized-errors), but still see this issue relatively frequently. I also received this email this morning:

image

Alternatively, we could just add that code to the sample, but I don't know what kind of security risk that creates for bots.


mdrichardson:
closed because it wasn't picking up my Teams commits

PR #2066, opened 2019-12-18
Add python teams Messaging Extension Action Preview

EricDahlvang:
Fixes https://github.com/microsoft/botbuilder-python/issues/482


samples/python/53.teams-messaging-extensions-action-preview/requirements.txt
```diff
@@ -0,0 +1,2 @@
+botbuilder-core>=4.4.0b1
+flask>=1.0.3
```
johnataylor: Neither of these requirements (aka dependencies seem correct) (i.e. should be botbuilder-core>=4.7.0 and aiohttp rather than flash.
How did you run this? Did you manually create your environment?

EricDahlvang: All of the samples in this branch are currently >=4.40b1

I did run this locally.  I'm not sure why it works, but I've updated the requirements.txt with aiohttp nonetheless

Thanks!

PR #2067, opened 2019-12-18
A complete MVC DirectLine token controller sample

ckkashyap:
Both https://blog.botframework.com/2018/09/25/enhanced-direct-line-authentication-features/ and https://blog.botframework.com/2018/09/01/using-webchat-with-azure-bot-services-authentication/ blogs contain snippets of how to create a controller to work with DirectLine Secure Site.

However, from customer feedback, these blogs could really use a complete sample, rather than just snippets of code. This PR is for such a sample in C#.


EricDahlvang:
This is defintely needed! But 36k files seems like a lot.  I think we could cut that down. 

 It would also be good to demonstrate multiple ways of doing this.

experimental/DirectLineTokenSite/Bot_Auth_DL_Secure_Site_MVC/Controllers/HomeController.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
```
johnataylor: all code should have the copyright header

ckkashyap:
> This is defintely needed! 36k lines in 53 files seems like a lot. I think we could cut that down.
> 
> It would also be good to demonstrate multiple ways of doing this.

Let's use this as a starting point in the experimental folder and we can build on this.

PR #2070, opened 2019-12-18
Simple Bot To Bot Skill sample (JS)

gabog:
Relates to https://github.com/microsoft/botframework-sdk/issues/5582

This commit contains a SimpleBotToBot example that shows the basics on how to consume an Echo Skill from a root bot.


samples/javascript_nodejs/70.skills-simple-bot-to-bot/echo-skill-bot/authentication/allowedCallersClaimsValidator.js
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+const { JwtTokenValidation, SkillValidation } = require('botframework-connector');
+
+// Load the AppIds for the configured callers (we will only allow responses from skills we have configured).
+// process.env.AllowedCallers is the list of parent bot Ids that are allowed to access the skill
+// to add a new parent bot simply go to the .env file and add
+// the parent bot's Microsoft AppId to the list under AllowedCallers, i.e:
```
stevengum: This is actually an `e.g.`, not an `i.e.` I'll fix that in another commit.

samples/javascript_nodejs/70.skills-simple-bot-to-bot/simple-root-bot/index.js
```diff
@@ -0,0 +1,101 @@
...
+const { RootBot } = require('./rootBot');
+
+// Import required bot configuration.
+const ENV_FILE = path.join(__dirname, '.env');
+require('dotenv').config({ path: ENV_FILE });
+
+// Create adapter.
+// See https://aka.ms/about-bot-adapter to learn more about adapters.
+const adapter = new BotFrameworkAdapter({
+    appId: process.env.MicrosoftAppId,
```
christopheranderson: suggestion: it's useful to emit an error if these environment variables aren't present. We see a lot of issues on come to Azure Bot Service related to these not being present and folks getting 401/etc.

stevengum: This is a little more nuanced as undefined/falsy values and empty strings are acceptable parameters here, and in this scenario. Bot-to-bot communication can be "anonymous". I'm going to mark it as out of scope because this would involve changing all the samples in all of the languages.

PR #2085, opened 2019-12-20
Python generator fixes

tracyboehrer:


generators/python/app/templates/echo/{{cookiecutter.bot_name}}/app.py
```diff
@@ -57,7 +57,7 @@
 ADAPTER.on_turn_error = on_error
 
 # Create the Bot
-BOT = EchoBot()
+BOT = MyBot()
```
EricDahlvang: This seems a little odd. (EchoBot named MyBot)

PR #2109, opened 2020-01-09
Dotnet: Send EndOfConversation to Skill in OnTurnError handler.

tracyboehrer:
Address 2101 for dotnet.

I did not resolve the active skill property name and it uses a magic string.  Since that error handler class is not terribly reusable, I didn't spend a lot of cycles on that aspect.  If you have a firm opinion on how you'd like it handled I'm open.

Also, there doesn't seem to be a good EndOfConversationCodes value.  I went with Unknown.

### DIFF
--- a/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
+++ b/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
@@ -26,5 +26,11 @@ namespace Microsoft.BotBuilderSamples.EchoSkillBot.Bots
                 await turnContext.SendActivityAsync(MessageFactory.Text("Say \"end\" or \"stop\" and I'll end the conversation and back to the parent."), cancellationToken);
             }
         }
+
+        protected override async Task OnEndOfConversationActivityAsync(ITurnContext<IEndOfConversationActivity> turnContext, CancellationToken cancellationToken)
+        {
+            var eocActivityMessage = $"Echo (dotnet): Received {ActivityTypes.EndOfConversation}, Code: {turnContext.Activity.Code}";
+            await turnContext.SendActivityAsync(MessageFactory.Text(eocActivityMessage), cancellationToken);
+        }
     }
 }
--- a/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
+++ b/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
@@ -33,7 +33,7 @@ namespace Microsoft.BotBuilderSamples.EchoSkillBot.Bots
             // This will be called if the root bot is ending the conversation.  Sending additional messages should be
             // avoided as the conversation may have been deleted.
             // Perform cleanup of resources if needed.
-            return Task.FromResult(true);
+            return Task.CompletedTask;
         }
     }
 }
--- a/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
+++ b/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Bot.Builder;
@@ -27,10 +28,12 @@ namespace Microsoft.BotBuilderSamples.EchoSkillBot.Bots
             }
         }
 
-        protected override async Task OnEndOfConversationActivityAsync(ITurnContext<IEndOfConversationActivity> turnContext, CancellationToken cancellationToken)
+        protected override Task OnEndOfConversationActivityAsync(ITurnContext<IEndOfConversationActivity> turnContext, CancellationToken cancellationToken)
         {
-            var eocActivityMessage = $"Echo (dotnet): Received {ActivityTypes.EndOfConversation}, Code: {turnContext.Activity.Code}";
-            await turnContext.SendActivityAsync(MessageFactory.Text(eocActivityMessage), cancellationToken);
+            // This will be called if the root bot is ending the conversation.  Sending additional messages should be
+            // avoided as the conversation may have been deleted.
+            // Perform cleanup of resources if needed.
+            return Task.FromResult(true);
         }
     }
 }
--- a/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
+++ b/samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
@@ -11,6 +11,7 @@ using Microsoft.Bot.Builder.Skills;
 using Microsoft.Bot.Builder.TraceExtensions;
 using Microsoft.Bot.Connector.Authentication;
 using Microsoft.Bot.Schema;
+using Microsoft.BotBuilderSamples.SimpleRootBot.Bots;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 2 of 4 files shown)


samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -17,57 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
-                    catch (Exception ex)
-                    {
-                        logger.LogError(ex, $"Exception caught on attempting to Delete ConversationState : {ex}");
-                    }
-                }
+            // Send a trace activity, which will be displayed in the Bot Framework Emulator
+            await turnContext.TraceActivityAsync("OnTurnError Trace", exception.ToString(), "https://www.botframework.com/schemas/error", "TurnError");
+        }
+
+        private async Task<InvokeResponse> endSkillConversationAsync(ITurnContext turnContext)
```
gabog: We don't seem to use the InvokeResponse value at all, this method should probably be `private async Task` so we don't return nulls or anything else that we don't use.

gabog: The method name should be PascalCase (EndSkillConversationAsync)

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -17,57 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
+            {
+                return null;
+            }
 
-                // Send a trace activity, which will be displayed in the Bot Framework Emulator
-                await turnContext.TraceActivityAsync("OnTurnError Trace", exception.ToString(), "https://www.botframework.com/schemas/error", "TurnError");
-            };
+            try
+            {
+                await _conversationState.SaveChangesAsync(turnContext, true);
```
gabog: Move the SaveChangesAsync right before the _skillClient.PostActivityAsync call, otherwise, the property you are creating below won't be persisted to state.

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -17,57 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
+
+                // Inform the active skill that the conversation is ended so that it has
+                // a chance to clean up.
+                // Note: "activeSkillProperty" is set by the RooBot while messages are being
+                // forwarded to a Skill.
+                var activeSkill = await _conversationState.CreateProperty<BotFrameworkSkill>("activeSkillProperty").GetAsync(turnContext, () => null);
```
gabog: Consider making "activeSkillProperty" a public const string in RootBot and update the code there too to use it.
It is still not awesome but at least we don't have the string hardcoded in two places.

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -17,57 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
+                return await _skillClient.PostActivityAsync(botId, activeSkill, _skillsConfig.SkillHostEndpoint, (Activity)endOfConversation, CancellationToken.None);
+            }
+            catch(Exception ex)
+            {
+                _logger.LogError(ex, $"Exception caught on attempting to send EndOfConversation : {ex}");
+                return null;
+            }
+        }
+
+        private async Task clearConversationStateAsync(ITurnContext turnContext)
```
gabog: Method name should be PascalCase

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/EchoSkillBot/Bots/EchoBot.cs
```diff
@@ -27,10 +28,12 @@ protected override async Task OnMessageActivityAsync(ITurnContext<IMessageActivi
...
 
-        protected override async Task OnEndOfConversationActivityAsync(ITurnContext<IEndOfConversationActivity> turnContext, CancellationToken cancellationToken)
+        protected override Task OnEndOfConversationActivityAsync(ITurnContext<IEndOfConversationActivity> turnContext, CancellationToken cancellationToken)
         {
-            var eocActivityMessage = $"Echo (dotnet): Received {ActivityTypes.EndOfConversation}, Code: {turnContext.Activity.Code}";
-            await turnContext.SendActivityAsync(MessageFactory.Text(eocActivityMessage), cancellationToken);
+            // This will be called if the root bot is ending the conversation.  Sending additional messages should be
+            // avoided as the conversation may have been deleted.
+            // Perform cleanup of resources if needed.
+            return Task.FromResult(true);
```
gabog: Return value should be Task.CompletedTask (the method is void, not bool).

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -13,42 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
+        private IConfiguration _configuration;
+        private ILogger _logger;
+        private SkillHttpClient _skillClient;
+        private SkillsConfiguration _skillsConfig;
+
+        public AdapterWithErrorHandler(IConfiguration configuration, ILogger<BotFrameworkHttpAdapter> logger, ConversationState conversationState = null, SkillHttpClient skillClient = null, SkillsConfiguration skillsConfig = null)
             : base(configuration, logger)
         {
-            OnTurnError = async (turnContext, exception) =>
+            _configuration = configuration;
```
gabog: Replace by the following to ensure we validate config is not null:
```
_configuration = configuration ?? throw new ArgumentNullException(nameof(configuration));
```

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -13,42 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
+            _skillClient = skillClient;
+            _skillsConfig = skillsConfig;
+
+            OnTurnError = OnBotError;
+        }
+
+        private async Task OnBotError(ITurnContext turnContext, Exception exception)
+        {
+            // Log any leaked exception from the application.
+            _logger.LogError(exception, $"[OnTurnError] unhandled error : {exception.Message}");
```
gabog: Replace by the following to ensure logger is not null:

``
_logger = logger ?? throw new ArgumentNullException(nameof(logger));;
``

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -13,42 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
             : base(configuration, logger)
         {
-            OnTurnError = async (turnContext, exception) =>
+            _configuration = configuration;
+            _conversationState = conversationState;
+            _logger = logger;
+            _skillClient = skillClient;
+            _skillsConfig = skillsConfig;
+
+            OnTurnError = OnBotError;
```
gabog: Should we called `HandleTurnError` instead of `OnBotError`?

samples/csharp_dotnetcore/80.skills-simple-bot-to-bot/SimpleRootBot/AdapterWithErrorHandler.cs
```diff
@@ -13,42 +18,96 @@ namespace Microsoft.BotBuilderSamples.SimpleRootBot
...
-                        await conversationState.DeleteAsync(turnContext);
-                    }
-                    catch (Exception ex)
-                    {
-                        logger.LogError(ex, $"Exception caught on attempting to Delete ConversationState : {ex}");
-                    }
-                }
+                var endOfConversation = Activity.CreateEndOfConversationActivity();
+                endOfConversation.Code = "RootSkillError";
+                endOfConversation.ApplyConversationReference(turnContext.Activity.GetConversationReference());
```
gabog: I don't think we need to apply the conversation reference here. SkillClient will do it for you.

tracyboehrer: I'll test.  On Python I had to otherwise the conversation was null and it threw an exception.

PR #2133, opened 2020-01-15
Python: Update SkillConversationId pattern

tracyboehrer:
Updated pattern for skill conversation ID to align with future considerations for OAuthPrompt.

Relates to https://github.com/microsoft/botframework-obi/issues/65


johnataylor:
can we just hold on this one for the moment - we are hoping we don't have to introduce micro-formats within IDs

johnataylor:
the current plan is that we won't be introducing the micro format in the conversation id field - so closing this PR

johnataylor:
It probably a good idea to keep this branch around for a little while because its possible this decision is reversed.

PR #2141, opened 2020-01-21
Update README.md

jiruss:
Fix CD command to correct path of repo

Fixes #

## Proposed Changes

  - Fixed the path for the CD command which incorrectly points to "Composer", should be BotFramework-Composer


jiruss:
Cancelling pull request. Realized this is asking for subfolder of Composer under the repo folder.

PR #2145, opened 2020-01-22
[Teams] Adding sample to creating new thread in Team (C#)

Virtual-Josh:


clearab:
@Virtual-Josh You can send the Activity with the initial createConversation call - you don't have to use continueConversation

clearab:
Err, I read that wrong, ignore me!

PR #2150, opened 2020-01-23
[Teams] Adding reply to channel sample for JS

Virtual-Josh:


Virtual-Josh:
> I think the README still needs to be updated, it looks like a copy/paste from the Conversation bot

Updated. readme didn't save previously.

clearab:
Made a couple more changes to the README

PR #2151, opened 2020-01-23
Moving QnA sample to the 70s to eliminate conflict

Virtual-Josh:


johnataylor:
Do we have to do this?
The number isn't very significant. A change like this can break [hypertext] links.

Virtual-Josh:
@johnataylor I spoke with Gurvinder and he's done with the PR and is fine with me moving the sample

PR #2158, opened 2020-01-30
Migrate samples to .Net Core 3.1

garypretty:
### C#

* Migrated samples to .Net Core 3.1
* Removed unnecessary usings across solution
* Updated adapter sample namespaces for consistency across solution

### JS

* use ActivityHandler.onEndOfConversation in JS samples (https://github.com/microsoft/BotBuilder-Samples/pull/2158/commits/140c8f879564e576231389e06e1ee181c68b9c33)


johnataylor:
Note this is in DRAFT state. We can't merge yet. This is for review. Please don't close this PR.

gabog:
I don't see a 4.7 branch in the code repo, shouldn't we create one based on what we released in 4.7.x before we merge this so we can point people to the 2.1 versions in the future in case they still need them?

garypretty:
@gabog Yes. This is not ready for merge yet - it's a draft PR. Its just for an initial review. 

It compiles but I need to test against Myget packages and we will only merge this once the 4.8 proper packages are available, before which we will branch samples.

gabog:
Makes sense, let's jut not forget about the branch before the release (we don't seem to be very consistent about release branches in the samples repo based on what I see)

gabog:
One more thing, shouldn't we update the samples in this branch to use 4.8.preview that is that version of the framework that has support for 3.1? (Skills won't work if we don't)

garypretty:
Yes, as I mentioned I will be updating to use 4.8 preview from myget. Just had some issues with it today. Should be done tomorrow. 👍

garypretty:
@gabog 👍

samples/csharp_dotnetcore/02.echo-bot/Startup.cs
```diff
@@ -35,22 +35,24 @@ public void ConfigureServices(IServiceCollection services)
...
-            {
-                app.UseHsts();
-            }
 
-            app.UseDefaultFiles();
-            app.UseStaticFiles();
-            app.UseWebSockets();
-            //app.UseHttpsRedirection();
-            app.UseMvc();
+            app.UseDefaultFiles()
```
gabog: The OOTB AspNet.Core template adds an app.UseEndpoints() call in startup.cs is that not needed?

Sample:
```
            app.UseEndpoints(endpoints =>
            {
                endpoints.MapControllers();
            });
```

samples/csharp_dotnetcore/02.echo-bot/Startup.cs
```diff
@@ -25,7 +25,7 @@ public Startup(IConfiguration configuration)
         // This method gets called by the runtime. Use this method to add services to the container.
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
+            services.AddControllers().AddNewtonsoftJson();
```
johnataylor: if we are being strict - this AddNewtonsoftJson() is not necessary here - is that correct?
it is only necessary if you were using skills and not using ExpectReplies

PR #2159, opened 2020-01-30
Fixed typo

rohannevrikar:
## Fixes 
Comment in line 59 should have "go through the cycle again" instead of "go though the cycle again".


sgellock:
@rohannevrikar thank you!

PR #2164, opened 2020-02-04
Handoffv2

arturl:
This set of changes implements feedback from today's design review

### DIFF
--- a/experimental/handoff-library/README.md
+++ b/experimental/handoff-library/README.md
@@ -6,11 +6,11 @@ The library implements the additions to the Bot Framework SDK to support handoff
 
 Integrations with specific agent hubs are not part of the library. 
 
-The library is intended to be merged in a future release of the Bot Framework SDK.
+The library is intto be merged in a future release of the Bot Framework SDK.
 
 ## Introduction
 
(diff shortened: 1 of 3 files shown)


EricDahlvang:
I see this removes the Transcript. As a developer, user or agent: some transcript option seems integral to a proper Handoff scenario.  Is that work planned for a future iteration? (curious)

arturl:
> I see this removes the Transcript. As a developer, user or agent: some transcript option seems integral to a proper Handoff scenario. Is that work planned for a future iteration? (curious)

@EricDahlvang, Transcript is not removed. See [here](https://github.com/microsoft/BotBuilder-Samples/tree/handoffv2/experimental/handoff-library#attachments).

EricDahlvang:
Oh, i see. Thanks. 

Are there plans to add paged transcript support to this api? Or is the expectation that the entire transcript is sent at the handoff initiation?

carlosscastro:
Great point @EricDahlvang on transcripts / attachments. We are having some discussions on that, I'll chat iwht you about that in person and get your thoughts :)

arturl:
@EricDahlvang, @carlosscastro - if the agent hub supports it, the bot can send transcript in "offline" attachments using the ContentUrl property. The protocol does not prohibit other schemes.

PR #2186, opened 2020-02-12
Fixed #2185

### LINKED ISSUE #2185: Sample 46 TeamsAuth: Logout not working if command is "logout"
## Sample information
1. Sample type: samples
3. Sample name: TeamsAuth

## Describe the bug
If a user types in command "logout" to sign out, the user does not get signed out. Instead the user is asked "Would you like to view a token". It works for commands having more words like "Please log me out" etc. 

## To Reproduce
1. Run the sample
2. Type anything to login
3. Type "logout"
4. It shows "Would you like to view your token".

## Expected behavior
If user types "logout", the bot should sign out user and show message "You have been signed out".

## Screenshots
image

[bug]

rohannevrikar:
This commit fixes the issue of "logout" command not working.

Fixes #2185 

- "logout" command is not working because the if condition is only true when text.IndexOf("logout") > 0, which would be false in case of text = "logout" as it'll return 0 index. It'll work for other commands like "please logout" etc. but not "logout". Hence, I changed the condition to text.IndexOf("logout") >= 0 so that "logout" command also works.

image

## Testing
Tested on web chat.

### DIFF
--- a/samples/csharp_dotnetcore/46.teams-auth/Dialogs/LogoutDialog.cs
+++ b/samples/csharp_dotnetcore/46.teams-auth/Dialogs/LogoutDialog.cs
@@ -48,7 +48,7 @@ namespace Microsoft.BotBuilderSamples
                 var text = innerDc.Context.Activity.Text.ToLowerInvariant();
 
                 // Allow logout anywhere in the command
-                if (text.IndexOf("logout") > 0)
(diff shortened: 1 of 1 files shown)


clearab:
This looks good to me, @johnataylor do you concur?

PR #2202, opened 2020-02-24
Update .NET generators for .NET Core 3.1

garypretty:
Update the .NET templates for .NET Core 3.1.

* Update templates to remove .NET Core 2.2 and replace with .NET Core 3.1
* Updated csproj package dependencies
* Update startup.cs / program.cs for each project
* Update all references to 4.6.2 / 4.7 to 4.8 in templates and visxmanifest

**Currently pointing to .NET 4.8-preview packages for testing purposes until 4.8 available - hence draft PR**


generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
```diff
@@ -45,16 +45,12 @@
         "datatype": "choice",
         "choices": [
           {
-            "choice": "netcoreapp2.2",
-            "description": "Target netcoreapp2.2"
-          },
-          {
-            "choice": "netcoreapp2.1",
-            "description": "Target netcoreapp2.1"
+            "choice": "netcoreapp3.1",
```
sgellock: I'm confused about what our plan is regarding the various supported versions of the .NET Core.  Based on the commit here, we *only* support 3.1.  This doesn't seem correct.  It seems we should support all supported .NET Core versions.  this would imply 2.2, no?

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
```diff
@@ -45,16 +45,12 @@
...
+            "description": "Target netcoreapp3.1"
           }
         ],
         "replaces": "__NETCOREAPP_VERSION__",
-        "defaultValue": "netcoreapp2.2"
+        "defaultValue": "netcoreapp3.1"
```
sgellock: Based on my feedback above, we can decide if 3.1 is the default or if we're going to ease into this.  Right now the command line arguments are incongruent with their intended purpose

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
```diff
@@ -64,39 +60,31 @@
...
           "evaluator": "C++",
           "cases": [
             {
-              "condition": "(Framework==\"netcoreapp2.2\")",
-              "value": "'Microsoft.AspNetCore' Version='2.2.0'"
-            },
-            {
-              "condition": "(Framework==\"netcoreapp2.1\")",
-              "value": "'Microsoft.AspNetCore' Version='2.1.6'"
+              "condition": "(Framework==\"netcoreapp3.1\")",
```
sgellock: Per above comments, removing the 2.2 runtime doesn't seem right

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot.Tests/CoreBot.Tests.csproj
```diff
@@ -21,8 +21,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.Bot.Builder.Testing" Version="4.7.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.1" />
+    <PackageReference Include="Microsoft.Bot.Builder.Testing" Version="4.8.0-preview-200219-106986" />
```
sgellock: This needs to be the final shipping assembly, not the preview assembly

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot.Tests/CoreBot.Tests.csproj
```diff
@@ -21,8 +21,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.Bot.Builder.Testing" Version="4.7.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.1" />
```
sgellock: I've gotten mixed feedback on this dependency.  Gabo had me add aspnetcore.app.  so I'd like to get a consistent set of feedback as to which is the appropriate asp.net assemblies to generate for the project.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot/CoreBot.csproj
```diff
@@ -14,10 +14,10 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.Bot.Builder.AI.Luis" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Bot.Builder.Dialogs" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Bot.Builder.Integration.AspNet.Core" Version="4.7.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.1" />
```
sgellock: Same issue as above.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot/CoreBot.csproj
```diff
@@ -14,10 +14,10 @@
...
+    <PackageReference Include="Microsoft.Bot.Builder.AI.Luis" Version="4.8.0-preview-200219-106986" />
```
sgellock: we need to use shipping assemblies

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot/Program.cs
```diff
@@ -13,16 +13,19 @@ public class Program
...
         {
-            CreateWebHostBuilder(args).Build().Run();
+            CreateHostBuilder(args).Build().Run();
         }
 
-        public static IWebHostBuilder CreateWebHostBuilder(string[] args) =>
-            WebHost.CreateDefaultBuilder(args)
-                .ConfigureLogging((logging) =>
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
```
sgellock: If these are 3.1 changes we're going to need to make changes to account for 2.2.  so this may make this more work than is currently completed.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot/Startup.cs
```diff
@@ -48,22 +48,24 @@ public void ConfigureServices(IServiceCollection services)
...
-                app.UseHsts();
-            }
 
-            app.UseDefaultFiles();
-            app.UseStaticFiles();
-            app.UseWebSockets();
-            //app.UseHttpsRedirection();
-            app.UseMvc();
+            app.UseDefaultFiles()
+                .UseStaticFiles()
```
sgellock: Again, if these are required for 3.1, then we're going to need to account for what works for 2.2 as well.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
garypretty: Our libs support both 2.2 and 3.1, as per the plan of record.  Our sample though, that consume the libs, need to target a version of .NET Core. The direction I was given from @cleemullins is that we should shift our samples to 3.1  We have also created a branch of the samples and marked it at 2.2 so that folks have a branch of those samples.

I think our only other option would be to either manage two copies of our samples which feels unmanageable or maybe we could add a couple of specific .NET 2.2 targeted samples such as for Core Bot?

Thoughts @sgellock?

sgellock:
I don’t agree with this direction.  It’s not the design goal I have put in place when I built these.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
garypretty: Sorry - I am clearly tired, hence I deleted my last comment which was talking about samples.

I think we have two choices - you are correct that we are supporting both 2.2 and 3.1.

1.  We have our templates target both 2.2 and 3.1 but this will cause complexity as we will need to have conditionals through the template in places.

2. Our SDK still supports both, but our templates target 3.1.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
garypretty: Agreed - this is going to depend on the outcome of my comment above.

sgellock:
I don't agree with this being changed to a 3.1 only template.  it violates a fundamental "meet the developer where they are" tenant that is core to what we do.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
garypretty: FWIW, I think supporting 2.2 is a good thing and necessary with the libs, but we don't have to continue to have our templates target both from my perspective.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot.Tests/CoreBot.Tests.csproj
garypretty: Yes - I have them on the preview assemblies right now until the final assemblies are published.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
sgellock: I have spoken with customers that have  use --framework to set the runtime version they are using.  this needs to be a design goal that we maintain.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot.Tests/CoreBot.Tests.csproj
garypretty: .NET Core 3.1 needs this dependency - they removed it from the core, so we need to add it. It isn't required for 2.2

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/CoreBot/Program.cs
garypretty: Agreed - if we want to provide the choice for both then we need to add conditionals throughout, but I am unsure of the benefits vs the cost here.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
garypretty: Understood - leave it with me.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
garypretty: As per above - leave it with me.

generators/dotnet-templates/Microsoft.BotFramework.CSharp.CoreBot/content/.template.config/template.json
sgellock: RE:  complexity of the templates - yes.  this will create complexity in our code but that complexity provides flexibility for our customers.  I'm willing to make this trade off

garypretty:
@sgellock We are aligned. Leave it with me.

garypretty:
@sgellock Closing this as we split the work up and agreed that this was the wrong approach anyway (excluding 2.1) - VISX updates now on #2238

PR #2231, opened 2020-03-09
Changed the Icon file names

brentos99:
The icon file names were icon-color.png icon-and outline.png, however in the manifest were missing the prefix icon-. This caused an error to occur when uploading the manifest.zip file to teams.


EricDahlvang:
Thanks @brentos99 !

PR #2246, opened 2020-03-17
[V3 Skill Sample] Catch exceptions during EndofConversation

EricDahlvang:
PVA sends EndOfConversation during manifest validation.  Since there is no active conversation, this will result in an exception.  This PR catches and ignores the exception.


EricDahlvang:
Determined the issue with PVA manifest validation is that they are sending an EndOfConversation activity with a null Recipient.

Closing this PR, as it is an ugly workaround...and the issue is actually with PVA.  Emailed them asking they include a Recipient in their eoc message.

PR #2253, opened 2020-03-18
Small fix for Sample 80 error handler

mdrichardson:
The error handler for the skill needs to stringify the error in the `text` value of the `endOfConversation` activity (it throws an error about it otherwise)

Note: Removed the conversationState part because there is no state in the skill

### DIFF
--- a/samples/javascript_nodejs/80.skills-simple-bot-to-bot/echo-skill-bot/index.js
+++ b/samples/javascript_nodejs/80.skills-simple-bot-to-bot/echo-skill-bot/index.js
@@ -63,9 +63,16 @@ adapter.onTurnError = async (context, error) => {
     const endOfConversation = {
         type: ActivityTypes.EndOfConversation,
         code: 'SkillError',
-        text: error
+        text: JSON.stringify(error)
     };
     await context.sendActivity(endOfConversation);
+
+    try {
+        // Clear out state
+        await conversationState.delete(context);
+    } catch (err) {
+        console.error(`\n [onTurnError] Exception caught on attempting to Delete ConversationState : ${ err }`);
+    }
 };
 
 // Create the main dialog.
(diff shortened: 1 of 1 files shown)


samples/javascript_nodejs/80.skills-simple-bot-to-bot/echo-skill-bot/index.js
```diff
@@ -63,9 +63,16 @@ adapter.onTurnError = async (context, error) => {
     const endOfConversation = {
         type: ActivityTypes.EndOfConversation,
         code: 'SkillError',
-        text: error
+        text: JSON.stringify(error)
```
stevengum: What does this look like from the parent side? Is the parent forwarding this text to the user or is this text being rendered in the console?

mdrichardson: Good ?. I'll do some testing and check. I just recall it throws an error in the Sample 81 I'm writing without this change. I wanted to make them match, but it might be handled differently--I'll check.

mdrichardson: I changed it from `JSON.stringify` to `toString`, which works better...it will actually show the error message instead of `{}`.

Here's the flow:

1. Error occurs in Skill
2. Error caught in Skill's error handler
2a. (without fix) error thrown in error handler because Error is not a string
2b. (with fix) Error is forwarded to Root as `text` in `endOfConversation` activity
4. Root displays the error in the chat because of:

```js
// Show status message, text and value returned by the skill
let eocActivityMessage = `Received ${ ActivityTypes.EndOfConversation }.\n\nCode: ${ context.activity.code }`;
if (context.activity.text) {
     eocActivityMessage += `\n\nText: ${ context.activity.text }`;
}

if (context.activity.value) {
    eocActivityMessage += `\n\nValue: ${ context.activity.value }`;
}
```

PR #2245, opened 2020-03-16
Skill Dialog Samples

gabog:
Fixes #2251
Fixes #2106
Fixes #2194

PR #2255, opened 2020-03-19
Python SkillDialog updates

axelsrz:
parity with #2245


axelsrz:
This code was merged into #2245 and will be reviewed there.

PR #2267, opened 2020-03-23
QnA: use DefaultAnswer config if passed in

KyleMit:
## Proposed Changes

The default QnA services chat bot initalizes with a value for `DefaultAnswer` in the azure configuration keys (implying that editing it would change the behavior).  Also, the documentation on how to [set default answer for a knowledge base](https://docs.microsoft.com/en-us/azure/cognitive-services/qnamaker/how-to/change-default-answer) indicates that you can set the value via the configuration key `DefaultAnswer` as well.

However, this sample project doesn't read in that value anywhere, and instead falls back to the constant [`DefaultNoAnswer `](https://github.com/microsoft/BotBuilder-Samples/blob/gary/4.8-dotnet-templates/samples/csharp_dotnetcore/49.qnamaker-all-features/Dialog/QnAMakerBaseDialog.cs#L23):

```cs
public const string DefaultNoAnswer = "No QnAMaker answers found.";
```

This update reads in the configuration value if it exists and falls back to the hard coded constant if it doesn't:

```diff
- noAnswer.Text = DefaultNoAnswer;
+ noAnswer.Text = this._configuration["DefaultAnswer"] ?? DefaultNoAnswer;
```

This issue came up in this question on [How can i set default answer in Q&A Azure bot](https://stackoverflow.com/a/60768373/1366033) as well

### DIFF
--- a/samples/csharp_dotnetcore/49.qnamaker-all-features/Dialog/QnAMakerBaseDialog.cs
+++ b/samples/csharp_dotnetcore/49.qnamaker-all-features/Dialog/QnAMakerBaseDialog.cs
@@ -7,6 +7,7 @@ using Microsoft.Bot.Builder.AI.QnA;
 using Microsoft.Bot.Builder.AI.QnA.Dialogs;
 using Microsoft.Bot.Builder.Dialogs;
 using Microsoft.Bot.Schema;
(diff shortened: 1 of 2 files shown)


cleemullins:
@KyleMit Thanks - great feature add to this sample!

PR #2282, opened 2020-03-27
Ckk/oauth support

ckkashyap:
An experimental sample to demonstrate the usage of ABS OAuth with BotKit adapter.


experimental/adapter-oauth/oauth_adapter.js
```diff
@@ -0,0 +1,134 @@
...
+            app_secret: fb_password, 
+            access_token: fb_access_token 
+       });
+
+       this.credentials = new botframework_connector_1.MicrosoftAppCredentials(appId, password);
+       this.credentialsProvider = new botframework_connector_1.SimpleCredentialProvider(appId, password);
+    }
+
+    getUserToken(context, connectionName, magicCode) {
+        return __awaiter(this, void 0, void 0, function* () {
```
benbrown, line 26: these _awaiter things are the result of this originally being copy pasted from compiled Typescript. We should do this with the normal async syntax and return the __awaiter stuff.

ckkashyap: Done

PR #2290, opened 2020-03-31
DotNet: Simplified 49.qnamaker-all-features.  README corrections.

tracyboehrer:
Simplified this sample, mainly because of a request from Doc.

QnAMakerBaseDialog really didn't do anything of value except redefine values from the real base class (QnAMakerDialog) with the same values.

RootDialog really did nothing more than fire up QnAMakerBaseDialog, which required the use of IBotServices and BotServices, none of which added much value to understanding the sample either.

Removed: RootDialog, IBotServices, BotServices.
Renamed: QnAMakerBaseDialog to BotQnAMakerDialog, and provided a constructor to give QnAMakerDialog the values it needs from config.

The README was confusing AF.  The Prerequisites led me down a path of darkness from which it was difficult to return.  I switched to using the steps from 11.qnamaker, which are a lot closer to the problem.  Though still not perfect (Doc is looking into this issue).


samples/csharp_dotnetcore/49.qnamaker-all-features/Dialog/BotQnAMakerDialog.cs
```diff
@@ -0,0 +1,36 @@
...
+    /// </summary>
+    public class BotQnAMakerDialog : QnAMakerDialog
+    {
+        public BotQnAMakerDialog(IConfiguration configuration)
+            :base(configuration["QnAKnowledgebaseId"], configuration["QnAEndpointKey"], BotQnAMakerDialog.GetHostname(configuration["QnAEndpointHostName"]))
+        {
+
+        }
+
+        private static string GetHostname(string hostname)
```
cleemullins, line 21: Please use the URI Class to do this. The code here is error prone, and should not be used. :) 

It will fail, for example with "httpS". :)

samples/csharp_dotnetcore/49.qnamaker-all-features/Dialog/BotQnAMakerDialog.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            if (!hostname.StartsWith("http"))
```
cleemullins, line 23: Instead of mutating, just throw an execption with a clear error message, saying the format of the string is wrong and should follow format XYZ.

tracyboehrer: I chose a safer way to make sure there is a scheme and path, then use Uri to validate, which will throw if not.

samples/csharp_dotnetcore/49.qnamaker-all-features/Dialog/BotQnAMakerDialog.cs
tracyboehrer: The intent of that method was to allow the user to specify either just a hostname or the full URL.  The Uri class does not do what is expected if someone doesn't specify a scheme.  Trying to introduce UriBuilder had similar gotchas.

mrivera-ms:
Hello @tracyboehrer  , this PR was closed accidentally during the process of renaming the branch 'master' to 'main'. Please open a new PR that references the new 'main' branch. I apologize for the inconvenience.

PR #2302, opened 2020-04-07
Move dialog generation to samples.

chrimc62:
This moves the experimental generation feature into samples.  This includes both the bf-generate plugin for doing generation and the RunBot utility for testing them.


experimental/generation/generator/src/commands/dialog/generate.ts
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                })
+            return true;
+        } catch (e) {
+            this.thrownError(e)
+        }
+    }
+
+    thrownError(err: Error): void {
+        this.error(err.message)
```
munozemilio, line 52: This will bypass the telemetry and errors wont be reported. It is better to throw the error instead

PR #2319, opened 2020-04-15
Adaptive tools refinement

Danieladu:
fix:

### Auto-completion 
-  when completing "-" for user, we need add an extra space "- "
- when directly hey enter without putting anything after "- " we should delete the "- "
- when completing functions we should complete with parameters as well or () if there is no parameters
-  when completing functions, if there is no conflicting with built-in functions, we should not add "lg." before functions, so that we can encourage a concise syntax
- it would be great to also indent after `-IF` and so on
### Validation
- seems like our validation not working on buffer, but only work when user saved the data


boydc2014:
How did you solve the perf issue?

and should we bump our version up a little bit?

PR #2372, opened 2020-05-07
Import problem

sahinyanlik:
Before changes it was returning channels => null

## Proposed Changes
Hello When I tried to run bot through emulator, in pictureStep it was returning undefined error. With this changes I used Channels enum in botbuilder-core


sgellock:
@sahinyanlik thank you for the PR.  greatly appreciated!

PR #2409, opened 2020-05-19
Linted & updated dependencies for echo and core bots

### LINKED ISSUE #2407: Linting 02.echo-bot and 13.core-bot to have yeoman templates and samples identical
For JS and TS samples, we want bots generated via yeoman template to match the bots that are in the samples.

Linting 02.echo-bot and 13.core-bot for R9, to keep them in sync.

Zerryth:
Fixes #2407

### Proposed Changes
Ultimately want yeoman templates and the bots in samples to make the same bot without discrepancies between the two.

*The following changes applied to both JS and TS versions of `02.echo-bot` and `13.core-bot` samples. These are the same changes done in yeoman generators [PR#2410](https://github.com/microsoft/BotBuilder-Samples/pull/2410)*:

> updates all samples to use 4.9.0 of Bot Framework SDK
updates all other nom dependencies to latest versions
updates typescript generated to to use TS 3.9.2
fix lint errors
update readme's to recommend emulator 4.9

#### Additional differences from generators vs. "general" echo and core bot samples that still exists
- For TS's core-bot, also made `run` method `public` as it needs its accessibility level explicitly declared
   - This change is not applied to generators, as core-bot in generators doesn't use `run` method
   - See comment for more details
- General node samples there's no JS empty bot sample (only in TS)

### DIFF
--- a/samples/javascript_nodejs/02.echo-bot/.eslintrc.js
+++ b/samples/javascript_nodejs/02.echo-bot/.eslintrc.js
@@ -1,3 +1,4 @@
+/* eslint-disable */
 module.exports = {
     "extends": "standard",
     "rules": {    
--- a/samples/javascript_nodejs/02.echo-bot/README.md
+++ b/samples/javascript_nodejs/02.echo-bot/README.md
@@ -43,7 +43,7 @@ This bot has been created using [Bot Framework](https://dev.botframework.com), i
 
 [Bot Framework Emulator](https://github.com/microsoft/botframework-emulator) is a desktop application that allows bot developers to test and debug their bots on localhost or running remotely through a tunnel.
 
-- Install the Bot Framework Emulator version 4.3.0 or greater from [here](https://github.com/Microsoft/BotFramework-Emulator/releases)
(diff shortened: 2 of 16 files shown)


Zerryth:
@sgellock 's working on the yeoman templates, and is definitely running the linter on the generators, which this PR does as well.

Making this only a draft PR, because there's a discrepancy between `13.core-bot` in the generators vs. what's in the general samples.

- *generators -- dialogBot.ts*
    - [Uses `onDialog` method](https://github.com/microsoft/BotBuilder-Samples/blob/sgellock/r9-yeoman/generators/generator-botbuilder/generators/app/templates/core/bots/dialogBot.ts#L47)

vs.

- *"general" bb-samples/samples/typescript_nodejs/13.core-bot ... dialogBot.ts*
    - [Uses `run` method](https://github.com/microsoft/BotBuilder-Samples/blob/master/samples/typescript_nodejs/13.core-bot/src/bots/dialogBot.ts#L51)

Zerryth:
@stevengum I applied the changes that Scott pushed yesterday to generators into "general" echo and core bot samples, but there are still discrepancies, detailed above.

Feel free to LMK if we want to handle eliminating those differences in R9 or wait until R10

PR #2410, opened 2020-05-19
Sgellock/r9 yeoman

sgellock:
## Proposed Changes
updates all samples to use 4.9.0 of Bot Framework SDK
updates all other nom dependencies to latest versions
updates typescript generated to to use TS 3.9.2
fix lint errors
update readme's to recommend emulator 4.9
update master readme.md to include steps to setup, test and run a local dev environment

## Testing
ran runTest.cmd without errors

### DIFF
--- a/generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
+++ b/generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
@@ -17,22 +17,22 @@
     },
     "dependencies": {
         "@microsoft/recognizers-text-data-types-timex-expression": "^1.1.4",
-        "botbuilder": "~4.7.0",
-        "botbuilder-ai": "~4.7.0",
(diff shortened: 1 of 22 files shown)


generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
```diff
@@ -17,22 +17,22 @@
     },
     "dependencies": {
         "@microsoft/recognizers-text-data-types-timex-expression": "^1.1.4",
```
stevengum, line 19: Not a diff in the PR, but this is the only recognizers-text dependency that's unpinned in the generators.

stevengum: Resolved via https://github.com/microsoft/BotBuilder-Samples/pull/2411 🚀

PR #2431, opened 2020-05-22
fix files without carriage return on the last line

flug:


cleemullins:
Hi @flug, I don't think we can take this PR. 

Before we took something like this, we would need a detailed issue, and a proposed solution. I'm not convinced this particular issue is a problem, but could be persuaded. 

Please open an issue and we can discuss.

PR #2456, opened 2020-05-29
Fix typo in qnamakerMultiturnDialog.js

v-kydela:
The typo is only present in Node


sgellock:
We've deleted this sample.

PR #2461, opened 2020-05-30
change template names for consistency.

tomlm:
current templates are confusing because
* it's not clear what the result of calling the template is, (text, activity, card, etc.)
* the template name has verbs which look like they do something but they dont' (example setPropertyMessage looks like a funciton that sets the property to a value, but it actually gets the description of setting the property. some templates act like boolean functions and some are generating content.

To make this clearer I have applied the following conventions:
# Prefix conventions:
* prefix with **get** for things that return something
* prefix with **is** for boolean templates
*etc.

# Postfix Conventions
* postfix with **Text** for things that return text.  This communicates to future editors that you shouldn't add a card in the middle of the template because others are expecting it to be text.
* postfix with **Card** for templates that return cards.
* postfix with **Activity** for templates that return activities.
* postfix with **Facts** for templates that return facts array
* postfix with **Fact** for templates that return afact
* etc.

# Examples
* **getPropertyValueText(property)** => gets the text for a property value.
* **getFormAsActivity()** => returns an activity .  It may be a text activity, or a card activity or a mixture, but the caller knows that it will return an activity.
* **getWelcomeText()** - returns text for welcoming.
* etc.\


experimental/generation/generator/templates/booleanEntity.en-us.lg.lg
```diff
@@ -5,7 +5,7 @@
 
 # template
 - ```
-${'#'} booleanEntity_DisplayValue(value)
+${'#'} getEntityValueText_booleanEntity(value)
```
chrimc62, line 8: >getEntityValueText_booleanEntity [](start = 7, length = 32)

You have put the property/entity last.  When we had talked before we had put that first so that you can find easily get completion on all the templates related to a property/entity.

experimental/generation/generator/templates/library-CompleteConfirmation.dialog.lg
```diff
@@ -10,11 +10,11 @@
     "actions": [
         {
             "$kind": "Microsoft.SendActivity",
-            "activity": "\${showForm('${prefix}', publicProperties())}"
+            "activity": "\${getFormAsActivity('${prefix}', getPublicPropertiesText())}"
```
chrimc62, line 13: >Form [](start = 31, length = 4)

Not sure if we should still keep form as a name, we don't use it anywhere else.  Not sure what would be better though.  'result'?  'schema'?

experimental/generation/generator/templates/library-Help.en-us.lg
```diff
@@ -1,32 +1,32 @@
 [library.en-us.lg]
 
-# askHelp
+# getAskHelpText
```
chrimc62, line 3: Maybe getAutoHelpText?

experimental/generation/generator/templates/library-Missing.en-us.lg
```diff
@@ -1,48 +1,48 @@
 [library.en-us.lg]
 [library-Help.en-us.lg]
 
-# askEnum(property)
+# getAskForEnumText(property)
```
chrimc62, line 4: >getAskForEnumText [](start = 2, length = 17)

Rather than Ask, thought we were using 'prompt'.

experimental/generation/generator/templates/library-CompleteConfirmation.dialog.lg
tomlm: dialog?


---
In reply to: [433372123](https://github.com/microsoft/BotBuilder-Samples/pull/2461#discussion_r433372123) [](ancestors = 433372123)

experimental/generation/generator/templates/library-Missing.en-us.lg
tomlm: Changed to 
getEnumPromptText

---
In reply to: [433372969](https://github.com/microsoft/BotBuilder-Samples/pull/2461#discussion_r433372969) [](ancestors = 433372969)

experimental/generation/generator/templates/library-Help.en-us.lg
tomlm: ok

---
In reply to: [433372587](https://github.com/microsoft/BotBuilder-Samples/pull/2461#discussion_r433372587) [](ancestors = 433372587)

experimental/generation/generator/templates/booleanEntity.en-us.lg.lg
tomlm: yeah, switched back.

---
In reply to: [433366141](https://github.com/microsoft/BotBuilder-Samples/pull/2461#discussion_r433366141) [](ancestors = 433366141)

experimental/generation/generator/templates/library-Assign.en-us.lg
```diff
@@ -1,75 +1,72 @@
 [library.en-us.lg]
 
-# numberValidation(property, number)
+# getNumberValidatationText(property, number)
```
chrimc62, line 3: >getNumberValidatationText [](start = 2, length = 25)

Mispelled.

experimental/generation/generator/templates/library-ShowForm.en-us.lg
```diff
@@ -1,18 +1,18 @@
...
+# getPropertiesAsText
+- ${join(foreach(dialog.getRequiredPropertiesText, property, getPropertyValueDescriptionText(property)), EOL())}
 
-# textProperty(property)
-- The value of ${displayName(property)} is ${displayValue(property, dialog[property])}.
+# getPropertyValueDescriptionText(property)
+- The value of ${getPropertyNameText(property)} is ${getPropertyValueText(property, dialog[property])}.
 
-# formCard(name, properties)
+# getPropertiesAsCard(name, properties)
```
chrimc62, line 15: >getPropertiesAsCard [](start = 2, length = 19)

Rename this file this is in to library-ShowProperties.en-us.lg?

experimental/generation/generator/templates/library-ShowFormIntent.dialog.lg
```diff
@@ -10,7 +10,7 @@
     "actions": [
         {
             "$kind": "Microsoft.SendActivity",
-            "activity": "\${showForm('${prefix}', dialogClass.schema.$public)}"           
+            "activity": "\${getPropertiesAsActivity('${prefix}', dialogClass.schema.$public)}"
```
chrimc62, line 13: Rename this file too.

PR #2464, opened 2020-06-02
Integrated LivePerson agent escalation into a Virutal Assistant Templ…

managedcoder:
This PR adds a new solution to the experimental/handoff-library/csharp_dotnetcore/samples folder that shows how to integrate LivePerson agent escalation into a Virutal Assistant Template bot using the new bot framework bot-as-a-proxy APIs for agent escalation.  

## Proposed Changes
This is an additional sample to the existing LivePersonProxyBot and its purpose is to show how to integrate the human handoff API in the Virtual Assistant Template which requires a different approach than what's shown in the LivePersonProxyBot.  Together, both samples give a full understanding of how the new human handoff APIs can be used to provide agent escalation. 

I had to make a change to the LivePersonConnector project to fit with how the Virtual Assistant's DefaultAdaptor is implemented.  That adapter has some important implementation in its constructor and so the change I made allowed LivePersonConnector to take an Adapter rather than force the consumer to use the Adaptor that it provided.  

Both LivePersonProxyBot and LivePersonProxyAssistant have their own LivePersonConnector projects but I've updated the LivePersonConnector in the LivePersonProxyBot with the new changes to make them both consistent.  I did have to included LivePersonCredentialsProvider.cs and LoggingConversationData.cs in the LivePersonConnector for LivePersonProxyAssistant since they were laying lose in LivePersonProxyBot but I did not reflect this change in LivePersonProxyBot to keep things as simple as possible.

BotInAction


experimental/handoff-library/csharp_dotnetcore/samples/LivePersonProxyAssistant/LivePersonConnector/ConversationMap.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using Microsoft.Bot.Schema;
+
+namespace LivePersonConnector
```
arturl, line 8: Can you make any necessary changes in the existing LivePersonConnector instead of cloning it?

PR #2474, opened 2020-06-04
Zoma/qna dir3

coldplaying42:
The command is :
bf dialog:generate schemaName -o . --templates experimental/generation/generator/templates/professional --templates experimental/generation/generator/templates/standard --singleton --qna --verbose 

To do cross-train, we need to add an intent into the file of en-us/schemaName.en-us.lu.


experimental/generation/generator/templates/professional/professional.schema
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "type": "object",
```
chrimc62, line 3: I don't think you need this.

experimental/generation/generator/test/commands/dialog/forms/sandwichQnA.schema
```diff
@@ -0,0 +1,102 @@
...
+        "Name",
+        "Bread",
+        "Cheese",
+        "Meat",
+        "Toppings",
+        "Sauces",
+        "Price"
+    ],
+    "$requires": [
+        "professional.schema",
```
chrimc62, line 99: This is exactly right for how you add this.

experimental/generation/generator/templates/standard/generator.lg
```diff
@@ -0,0 +1,359 @@
+> This file contains templates to help with the generation process
```
chrimc62, line 1: Probably move this file to the parent directory and change all the references to point to it and add what you need for qna.  It looks like the one in professional is pretty much the same as this one.

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -530,6 +531,22 @@ async function generateSingleton(schema: string, inDir: string, outDir: string)
...
+                let content = await fs.readFile(ppath.join(inDir, file), 'utf-8')
+                qnaText = qnaText + os.EOL + content.toString().replace(ReplaceGeneratorPattern, '')
+                await fs.unlink(ppath.join(inDir, file))
+            }
+        }
+        let qnaName = `${schema}.${locale}.qna`
+        qnaText += `${os.EOL}> Generator: ${computeHash(qnaText)}`
+        await fs.writeFile(ppath.join(`${inDir}/${locale}`, qnaName), qnaText)
+    }
+}
```
chrimc62, line 549: Maybe I'm missing something, but I don't think you need this at all.  It should all be done through templates.

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -644,7 +662,7 @@ export async function generate(
             await fs.emptyDir(outPathSingle)
         }
 
-        let standard = normalize(ppath.join(__dirname, '../../templates'))
+        let standard = normalize(ppath.join(__dirname, '../../templates/standard'))
         templateDirs = expandStandard(templateDirs)
```
chrimc62, line 666: We can drop expandStandard.  They way it should work:
1) We join specified template dirs + standard and professional and use these to find schemas.  This makes it easy to refer to standard.schema, professional.schema or other stuff we add.  
2) For looking for templates we use template dirs in the order specified plus any dirs with schema in the order was used. (That is what the code below after finding schema does.)

chrimc62:
> 

Remove this file and put anything new you need in generator.lg in the parent templates directory.

---
Refers to: experimental/generation/generator/templates/professional/generator.lg:2 in 11515f2. [](commit_id = 11515f2f191b75bcb707d820018298af47b35097, deletion_comment = False)

experimental/generation/generator/templates/professional/build.cmd.lg
```diff
@@ -0,0 +1,34 @@
+# filename
+-build.cmd
```
chrimc62, line 2: If you are overriding this file, you also need to override build.lg which has the bash script as well.

experimental/generation/generator/templates/professional/professional.schema
```diff
@@ -0,0 +1,18 @@
...
+    "$templates": [
+        "AboutBot.qna",
+        "AboutUser.qna",
+        "Command.qna",
+        "Compliment.qna",
+        "Criticism.qna",
+        "Dialog.qna",
+        "Greeting.qna",
+        "Relationship.qna",
+        "library-CrossTrain.lg",
```
chrimc62, line 13: You can get rid of your qna function if you add a locale.qna.lg file that is like the locale.lg.lg or locale.lu.lg file in standard.

experimental/generation/generator/templates/professional/library-ChooseIntent.dialog.lg
```diff
@@ -0,0 +1,59 @@
...
+- ```
+{
+    "$schema": "${appSchema}",
+    "$kind": "Microsoft.OnChooseIntent",
+    "actions": [
+        {
+            "$kind": "Microsoft.SetProperty",
+            "property": "dialog.recognized",
+            "value": "=turn.recognized"
+        },
```
chrimc62, line 11: Why is this needed?

experimental/generation/generator/templates/professional/library-ChooseIntent.dialog.lg
```diff
@@ -0,0 +1,59 @@
...
+        {
+            "$kind": "Microsoft.ChoiceInput",
```
chrimc62, line 13: >"$kind": "Microsoft.ChoiceInput", [](start = 12, length = 33)

This really should be an ask--we want open-ended prompts.  Replacing it is more complex, so we can leave it for now but file a bug to come back and fix it.

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -141,7 +141,7 @@ const expressionEngine = new expressions.ExpressionParser((func: string): any =>
     }
 })
 
-const generatorTemplate = lg.Templates.parseFile(ppath.join(__dirname, '../../templates', 'generator.lg'), undefined, expressionEngine)
+const generatorTemplate = lg.Templates.parseFile(ppath.join(__dirname, '../../templates/standard', 'generator.lg'), undefined, expressionEngine)
```
chrimc62, line 145: Keep the original location for generator.lg

chrimc62:
>         } else {

Should add a break here as well.

---
Refers to: experimental/generation/generator/src/library/dialogGenerator.ts:179 in 11515f2. [](commit_id = 11515f2f191b75bcb707d820018298af47b35097, deletion_comment = False)

chrimc62:
> function expandStandard(dirs: string[]): string[] {

Remove this function.  See below for why.

---
Refers to: experimental/generation/generator/src/library/dialogGenerator.ts:561 in 11515f2. [](commit_id = 11515f2f191b75bcb707d820018298af47b35097, deletion_comment = False)

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -578,6 +595,7 @@ export async function generate(
     force?: boolean,
     merge?: boolean,
     singleton?: boolean,
+    qna?: boolean,
```
chrimc62, line 598: Remove this flag.

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -691,6 +709,11 @@ export async function generate(
             }
         }
 
+        if (qna) {
+            feedback(FeedbackType.info, 'Merging qna files')
+            await generateQnA(scope.prefix, allLocales, outDir)
+        }
```
chrimc62, line 715: Remove this.

experimental/generation/generator/src/commands/dialog/generate.ts
```diff
@@ -26,7 +26,8 @@ export default class GenerateDialog extends Command {
         output: flags.string({ char: 'o', description: 'Output path for where to put generated .lu, .lg, .qna and .dialog files.', default: '.', required: false }),
         prefix: flags.string({ char: 'p', description: 'Prefix to use for generated files. [default: schema name]' }),
         schema: flags.string({ char: 's', description: 'Path to your app.schema file.', required: false }),
-        singleton: flags.boolean({ description: 'Specify to generate a single .dialog file.'}),
+        singleton: flags.boolean({ description: 'Specify to generate a single .dialog file.' }),
+        qna: flags.boolean({ description: 'Specify to include qna' }),
         templates: flags.string({ char: 't', description: 'Directory with templates to use for generating assets.  With multiple directories, the first definition found wins.  To include the standard templates, just use "standard" as a template directory name.', multiple: true }),
```
chrimc62, line 31: Drop this flag.

experimental/generation/generator/templates/professional/professional.schema
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
```
chrimc62, line 2: Maybe rename this file professional-chit-chat.schema?

experimental/generation/generator/templates/professional/library-ChooseIntent.dialog.lg
coldplaying42: When the utterance is recognized by luis and qnamaker with the fair score, we need the user to choose intent.

chrimc62: I see this is persisting so you can use it for the final result.  
To switch this to an ask.  (Not needed now.)
1) Have this trigger set dialog.chooseRecognizer=turn.recognized
2) Have a separate trigger that is triggered by the presence of dialog.chooseRecognizer and has a 0 priority.  It should do the ask.
3) Setup the LUIS models to detect the different sources and a way to say none of them.
4) When one of the selections happens, have a trigger which selects the appropriate one and clear dialog.chooseIntent

---
In reply to: [435744407](https://github.com/microsoft/BotBuilder-Samples/pull/2474#discussion_r435744407) [](ancestors = 435744407)

experimental/generation/generator/templates/professional/build.cmd.lg
coldplaying42: Does this mean in the build.cmd.lg file, we have an if-else clause. If it has qna file, return the template with luis:build/cross-train, or else, return the template with luis:build ?

experimental/generation/generator/src/library/dialogGenerator.ts
coldplaying42: All I need to do is just deleting the expandStandard function ?

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -275,6 +277,8 @@ async function processTemplate(
                     } else if (filename.includes('library-')) {
                         // Put library stuff in its own folder by default
                         filename = `library/${filename}`
+                    } else if (filename.includes('qna')) {
+                        filename = `${scope.locale}/${filename}`
                     }
```
chrimc62, line 282: I think this is already covered by the filename including the locale right?  Why would we need for just qna and not lu/lg?

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -288,7 +292,7 @@ async function processTemplate(
                         // This is a new file
                         if (force || !await fs.pathExists(outPath)) {
                             feedback(FeedbackType.info, `Generating ${outPath}`)
-                            let result = plainTemplate?.template
+                            let result = plainTemplate ?.template
```
chrimc62, line 295: >e ? [](start = 53, length = 3)

Not sure, but I thought our style did not have a space on this?

experimental/generation/generator/src/library/dialogGenerator.ts
chrimc62: Delete the function and this call.  You also need to ensure all our standard templates are included when we process schemas, i.e. you want the below so you can find all schemas we define.
<passedIn templates>, templates/standard, templates/professional.  (Bonus point for just enumerating templates directory to generate this.)  
Once you processSchemas, you change to <passedIn templates> + <schema template directories>, i.e. what is passed in plus template directories associated with schemas people use.

---
In reply to: [436688418](https://github.com/microsoft/BotBuilder-Samples/pull/2474#discussion_r436688418) [](ancestors = 436688418)

experimental/generation/generator/templates/professional/professional.schema
chrimc62: Still don't need this.

---
In reply to: [435642984](https://github.com/microsoft/BotBuilder-Samples/pull/2474#discussion_r435642984) [](ancestors = 435642984)

experimental/generation/generator/templates/professional/Relationship.qna
```diff
@@ -0,0 +1,99 @@
+> Collection of chit chat interactions between user and bot. You can add additional questions or adapt the answer as you see fit. Here's the source editorial dataset - [Source](https://github.com/Microsoft/BotBuilder-PersonalityChat/blob/master/CSharp/Datasets/Queries_Responses_Professional_QnAMaker.tsv)
+
```
chrimc62, line 2: All of the .qna files should be X.en-us.qna--then you don't need the special case any more. They should also be in an en-us subdirectory.  There are 5 styles and 9 languages for .qna we will bring in here soon as well as If you haven't used it the new powertoys has a nice renaming tool.

experimental/generation/generator/templates/professional/build.cmd.lg
chrimc62: Yes.  You can either use a function if or a template one, but the idea is the only difference is cross-train/build so capture that difference.

---
In reply to: [436673619](https://github.com/microsoft/BotBuilder-Samples/pull/2474#discussion_r436673619) [](ancestors = 436673619)

experimental/generation/generator/src/library/dialogGenerator.ts
chrimc62: I don't think we need to do anything special for .qna. Just put it in the flat locale directory.

---
In reply to: [436976226](https://github.com/microsoft/BotBuilder-Samples/pull/2474#discussion_r436976226) [](ancestors = 436976226)

experimental/generation/generator/templates/standard/build.cmd.lg
```diff
@@ -0,0 +1,63 @@
...
+set schemaName=%7
+
+echo Building Models
+bf luis:cross-train --config ./config.json --in . && copy .\cross-trained\%schemaName%.%locale%.lu .\%locale%\ /y && bf luis:build --luConfig luconfig.json --region=%region% %key% --suffix %alias% --botName %botName% --out . --log --force && bf qnamaker:build --subscriptionKey %qnaKey% --in .\cross-trained\ --suffix %alias% --botName %botName% --out . --log --force 
+
+goto done
+
+:help
+echo build.cmd [region] [locale] [authoringKey] [qnaKey] [alias] [botName] [schemaName]
+echo Region defaults to westus.
```
chrimc62, line 61: You don't need to make the whole thing conditional--only the bf line and the documentation

experimental/generation/generator/templates/standard/build.lg
```diff
@@ -0,0 +1,67 @@
...
+set schemaName=$7
+
+bf luis:cross-train --config ./config.json --in . && copy .\cross-trained\$schemaName.$locale.lu .\$locale\ /y && bf luis:build --luConfig $thisdir/luconfig.json --region=$region $key --out $thisdir --suffix $alias --botName $botName --out . --log --force && bf qnamaker:build --subscriptionKey $qnaKey --in .\cross-trained\ --suffix $alias --botName $botName --out . --log --force 
+res=$?
+
+if [ $res != 0 ]; then
+echo build.cmd [region] [locale] [authoringKey] [qnaKey] [alias] [botName] [schemaName]
+echo Region defaults to westus.
+fi
+
```
chrimc62, line 65: Same here.  Only make conditional the changed parts.

experimental/generation/generator/src/library/dialogGenerator.ts
```diff
@@ -257,21 +259,23 @@ async function processTemplate(
...
                         } catch (e) {
                             throw new Error(`${templateName}: ${e.message}`)
                         }
-                    } else if (filename.includes(scope.locale)) {
+                    } else if (filename.includes(scope.locale) && !filename.includes('qna')) {
                         // Move constant files into locale specific directories
                         let prop = templateName.startsWith('library') ? 'library' : scope.property
                         filename = `${scope.locale}/${prop}/${filename}`
+                    } else if (filename.includes(scope.locale)) {
+                        filename = `${scope.locale}/QnA/${filename}`
```
chrimc62, line 278: This should still be removed and the !filename.includes('qna') from the previous clause.  If you want the prop folder to be QnA put the test inside setting prop.

PR #2475, opened 2020-06-04
Remove the root (Component) dialog from multi-turn prompt sample

JonathanFingold:
I can't figure out why the root component dialog is part of this sample, other than tradition.

I've removed the root dialog and tested that the behavior is the same without it.


cleemullins:
@JonathanFingold Vishwac has approved this. Should we remove the "Do not merge" tag and merge it in?

JonathanFingold:
@cleemullins @vishwacsena This PR was more a proof of concept that a polished refactoring. :| I can clean this up the rest of the way in a new PR as a proposal for a basic pattern for our adaptive dialog samples.

PR #2484, opened 2020-06-05
Add multi-language prompt sample with language generation

### LINKED ISSUE #2481: Add Node.js sample for Multi-lingual generation and language fallback policy
The section [Multi-lingual generation and language fallback policy](https://docs.microsoft.com/en-us/azure/bot-service/bot-builder-concept-language-generation?view=azure-bot-service-4.0&tabs=javascript#multi-lingual-generation-and-language-fallback-policy) only has an sample for dotnet. I think that the docs would really benefit from an example in Node.js. If others are interested I got it working after some trial and error.

In your Bot and/or Dialog file (or other files where you use LG) add:

`const { Templates, MultiLanguageLG } = require('botbuilder-lg');`

In your constructor add:

```
// Create Templates for all locales (must use absolute path)
const defaultLg = Templates.parseFile(__dirname + '/lg/Main.lg'); // default is english in my case
const svLg = Templates.parseFile(__dirname + '/lg/Main.sv-SE.lg'); // swedish translations

// Create a Map with Templates for all locales
const lgTemplatesPerLocale = new Map<string, string>();
lgTemplatesPerLocale.set("", defaultLg);
lgTemplatesPerLocale.set("sv-SE", svLg); // Note, language code must match exactly what your client is using!

// Create a Multi Language LG using the Templates in the Map
this.lgGenerator = new MultiLanguageLG(lgTemplatesPerLocale);
```
The constructor in `MultiLanguageLG` can take a Map with file path to the LG file as a string (filePerLocale), instead of a Tampletes object, but I never managed to get it working. I am unsure if absolute or relative file paths must be used. The reason i use `Templates.parseFile` is that an error shows in the console if the path is not correct (the lg file can not be found). This helps making sure the LG files are found or not. No error messages are printed when using a filePerLocale Map.

So, doing like this is not working:

```
const lgFilesPerLocale= new Map<string, string>();
lgFilesPerLocale.set("", __dirname + '/lg/Main.lg');
lgFilesPerLocale.set("sv-SE", __dirname + '/lg/Main.sv-SE.lg');
this.lgGenerator = new MultiLanguageLG(lgFilesPerLocale);
```
No useful error message is displayed in console, but:

```
(node:22396) UnhandledPromiseRejectionWarning: TypeError: Cannot perform 'set' on a proxy that has been revoked
    at adapter.sendActivities.then (D:\Utveckling\Teams Apps\moodle-stats\node_modules\botbuilder-core\lib\turnContext.js:339:36)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:22396) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a
catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:22396) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Next step is to call `generate` on the `lgGenerator` when you want your localized messages:

`this.lgGenerator.generate('YourTemplateName', yourDataToTheTemplate, context.activity.locale);`

If you do not need to send data to the template use `""` (using `null` did not work, generated an error message in console):

`this.lgGenerator.generate('YourTemplateName', "", context.activity.locale);`

Exampel of sending an Activity:

`await context.sendActivity(ActivityFactory.fromObject(this.lgGenerator.generate("WelcomeCard", userProfile, context.activity.locale)));`

Note! Be sure to check what `context.activity.locale` contains and then you **must match exactly** this value in the Map. For example in my MS Teams `context.activity.locale` contains `sv-SE`. If I in the map use:

`lgTemplatesPerLocale.set("sv", svLg);`

It will not work. Neither will:

`lgTemplatesPerLocale.set("sv-se", svLg);`

Notice the difference in se and SE. This is strange as I thoug language tags would be handled in the same way as localization for the manifest.json file (localizationInfo). At the page, linked below, they _"strongly recommended to provide top-level, language-only translations in your manifest ('en' instead of 'en-us') and only provide region-level overrides for the few strings that need them"_.

https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/build-and-test/apps-localization#localizing-the-strings-in-your-app-manifest

[enhancement]

vishwacsena:
Fixes #2481


Danieladu:
LGTM. Seems no problem in LG file, and the bot reply is expected for multi-lingual

PR #2485, opened 2020-06-06
Live person proxy assistant factored

managedcoder:
This PR adds a new solution to the experimental/handoff-library/csharp_dotnetcore/samples folder that shows how to integrate LivePerson agent escalation into a Virutal Assistant Template bot using the new bot framework bot-as-a-proxy APIs for agent escalation.

## Proposed Changes
This is an additional sample to the existing LivePersonProxyBot and its purpose is to show how to integrate the human handoff API in the Virtual Assistant Template which requires a different approach than what's shown in the LivePersonProxyBot. Together, both samples give a full understanding of how the new human handoff APIs can be used to provide agent escalation.

I had to make a change to the LivePersonConnector project to fit with how the Virtual Assistant's DefaultAdaptor is implemented. That adapter has some important implementation in its constructor and so the change I made allowed LivePersonConnector to take the DefaultAdaptor from the Virtual Assistant Template rather than force the consumer to use the LivePersonAdaptor.

To accommodate the Virtual Assistant Template I had to refactor several things which ultimately simplified LivePersonProxyBot and made integration with the Virtual Assistant Template possible.  Most notably, I pulled LivePersonConnector out of the LivePersonProxyBot sample and now that single library is shared between the two samples. 

I also moved LivePersonCredentialsProvider.cs and LoggingConversationData.cs into the LivePersonConnector since LivePersonProxyAssistant needed them too.

Both samples were pulled from the repo fresh and verified to be working and here are the screen grabs of both LivePersonProxyBot and LivePersonPrixyAssistant working:

## LivePersonProxyBot working:
image

## LivePersonProxyAssistant
image


experimental/handoff-library/csharp_dotnetcore/samples/LivePersonConnector/Middleware/LoggingMiddleware.cs
```diff
@@ -24,10 +24,10 @@ public LoggingMiddleware(ConversationState conversationState)
 
         public async Task OnTurnAsync(ITurnContext turnContext, NextDelegate next, CancellationToken cancellationToken = default)
         {
-            // Route the conversation based on whether it's been escalated
```
arturl, line 27: thanks! :-)

PR #2518, opened 2020-06-19
update scripts

coldplaying42:
Update config, run and build lg files to support qna and cross-train.

### DIFF
--- a/experimental/generation/generator/templates/standard/build.cmd.lg
+++ b/experimental/generation/generator/templates/standard/build.cmd.lg
@@ -1,3 +1,5 @@
+[generator.lg](../generator.lg)
+
 # filename
 -build.cmd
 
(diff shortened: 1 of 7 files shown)


experimental/generation/generator/templates/standard/build.cmd.lg
```diff
@@ -41,7 +40,13 @@ goto done
...
 call bf luis:cross-train --config ./config.json --in . 
-call copy .\cross-trained\%schemaName%.%locale%.lu .\%locale%\ /y 
+${join(foreach(locales, locale, copyForLocale(locale)) , EOL())}
+```
+
+# copyForLocale(locale)
+-```
+call copy .\cross-trained\\${prefix}.${locale}.lu .\\${locale}\ /y
+call copy .\cross-trained\\${prefix}.${locale}.qna .\\${locale}\ /y
 ```
```
chrimc62, line 50: Couldn't you just put the files in the right place to start with?

coldplaying42: That is because when we do bf luis:cross-train --config ./config.json --in .  ,  config.json  might have multi-language lu (e.g, en-us/sandwich.en-us.lu and zh-ch/sandwich.zh-ch.lu) and its corresponding qna files. After cross-train, all the updated lu  (e.g., sandwich.en-us.lu and sandwich.zh-ch.lu ) and qna files will be stored in the cross-trained folder, so we need to copy them back to their corresponding  locale folder.

PR #2521, opened 2020-06-21
Update RootDialog.lu

irwinwilliams:
There doesn't seem to be a Greeting intent in the lu file.

Fixes #

## Proposed Changes
Literally adding the Greeting intent to the Root lu file.


cleemullins:
@vishwacsena Please chime in here with your opinion.

PR #2522, opened 2020-06-22
T sheny/completion lg

ShYuPe:
## Description
A vscode extension implementation for language-generation completion.

Options checklist

- Strict option
strict

- replaceNull option 
replaceNull

- lineBreakStyple option 
lineBreakStyple

- Namespace option 
namespace

- Exports option 
exports


experimental/adaptive-tool/src/lg/providers/completion.ts
```diff
@@ -91,6 +91,96 @@ class LGCompletionItemProvider implements vscode.CompletionItemProvider {
                 })
                 return items;
             }
+        } else if (/^> !#/.test(lineTextBefore)) {
```
Danieladu, line 94: It's better to have some comments about this block to show waht does "/^> !#/.test(lineTextBefore)" mean.

experimental/adaptive-tool/src/lg/providers/completion.ts
```diff
@@ -91,6 +91,96 @@ class LGCompletionItemProvider implements vscode.CompletionItemProvider {
...
+            let items: vscode.CompletionItem[] = [];
+            if (/> !#$/.test(lineTextBefore)) {
+                var options : { [key: string] : { [key: string] : string; }; } = { 
+                    "@strict": {
```
Danieladu, line 98: Maybe the options could be optimized to const data out of the code block.

experimental/adaptive-tool/src/lg/providers/completion.ts
```diff
@@ -91,6 +91,96 @@ class LGCompletionItemProvider implements vscode.CompletionItemProvider {
...
+                    };
+                for (let option in options) {
+                    let completionItem = new vscode.CompletionItem(" " + option);
+                    completionItem.detail = options[option]["detail"];
+                    completionItem.documentation = options[option]["documentation"];
+                    completionItem.insertText = options[option]["insertText"];
+                    items.push(completionItem);
+                }
+            } else {
+                if (/> !# *@strict *=$/.test(lineTextBefore)) {
```
Danieladu, line 127: use `\s` instead of ' '

experimental/adaptive-tool/src/lg/providers/completion.ts
```diff
@@ -91,6 +91,96 @@ class LGCompletionItemProvider implements vscode.CompletionItemProvider {
...
+                    var strictOptions : { [key: string] : { [key: string] : string; }; } = {
+                        "true": {
```
Danieladu, line 129: extract it into data

Danieladu:
Maybe you can investigate some solutions that can reduce repeated code patterns.
Something like:
```
var optionsMap = [
 {
   strict: [
   {}//complition list
   ],
  Namespace: [
   //xxx
  ]
 }
]

for(var item in optionsMap ) {
 if (match(item.key)){
// add item.value into completion
}
}
```

experimental/adaptive-tool/src/lg/providers/completion.ts
```diff
@@ -141,6 +165,70 @@ class LGCompletionItemProvider implements vscode.CompletionItemProvider {
...
+            let completionItem = new vscode.CompletionItem(" " + option);
+            completionItem.detail = options[option]["detail"];
+            completionItem.documentation = options[option]["documentation"];
+            completionItem.insertText = options[option]["insertText"];
+            items.push(completionItem);
+        }
+    }
+}
 
+const optionsMap : { [key: string] : { [key: string] : { [key: string] : string; }; }; } = {
```
Danieladu, line 180: there is no need to construct a strong type, a simple object is enough.
```
const optionsMap : { [key: string] : { [key: string] : { [key: string] : string; }; }; } = {"a": {"b":"c"}}
- >
 const optionsMap  = {a: {b: 'c'}}
```

So the scope can be instead as:
```
const optionsMap = {
    options: { 
        "@strict": {
            detail: " @strict = true",
            documentation: "Developers who do not want to allow a null evaluated result can implement the strict option.", 
            insertText: " @strict"},
      xxxx
    },
    strictOptions : {
        true: {
            detail: " true",
            documentation: "Null error will throw a friendly message.",
            insertText: " true"
        },
        false: {
            detail: " false",
            documentation: "A compatible result will be given.",
            insertText: " false"
        }
    },
    xxxx
}
```
This is more readable

Danieladu:
another tip: install eslint extension to fix all syntax warnings and errors.

experimental/adaptive-tool/src/lg/providers/completion.ts
```diff
@@ -91,6 +91,30 @@ class LGCompletionItemProvider implements vscode.CompletionItemProvider {
                 })
                 return items;
             }
+        } else if (/^>\s!#/.test(lineTextBefore)) {
+            // options suggestion following "> !#"
+            let items: vscode.CompletionItem[] = [];
+            if (/>\s!#$/.test(lineTextBefore)) {
+                this.AddToCompletion(optionsMap["options"], items);
```
Danieladu, line 98: optionsMap.options is ok

PR #2537, opened 2020-06-24
Fix a mistake in html that using a wrong unit of font size.

Anduin2017:
Fix a mistake in HTML that using a wrong unit of the font size.

## Proposed Changes

  -Fix a mistake in HTML that using a wrong unit of the font size.


tracyboehrer:
@Anduin2017 Hi Adnduin, this PR is in conflict with master.  Can you resolve the conflict?

PR #2538, opened 2020-06-24
fully description of TIMEX in readme

### LINKED ISSUE #1471: Docs on the timex results from LUIS are sparse
The datetimev2 / timex extractor feature of LUIS is super cool and powerful and works really well.

However, the docs on how to use timex are very minimal, and currently point to a very academic PDF document.

It would be great if we could provide examples of using the timex to do normal things like creating a single or repeating calendar schedule type date from the timex.

A simple breakdown of the various formats used by Timex and what they mean would also be very welcome!

Current docs:
https://docs.microsoft.com/en-us/azure/cognitive-services/luis/luis-reference-prebuilt-datetimev2

Timex explanation:
http://www.timeml.org/tempeval2/tempeval2-trial/guidelines/timex3guidelines-072009.pdf

johnataylor:
Fixes #1471

### DIFF
--- a/samples/csharp_dotnetcore/40.timex-resolution/README.md
+++ b/samples/csharp_dotnetcore/40.timex-resolution/README.md
@@ -2,6 +2,20 @@
 
 This bot has been created using [Bot Framework](https://dev.botframework.com), is shows how to use TIMEX expressions.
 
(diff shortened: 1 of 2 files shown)


johnataylor:
fixes #2538 1471

PR #2629, opened 2020-07-28
TS teams messaging extensions search sample

stevkan:
## Proposed Changes
Adds TS 50.teams-messaging-extensions-search sample


tracyboehrer:
This appears to work in Teams according to the README

PR #2630, opened 2020-07-28
TS Teams messaging extensions action sample

stevkan:
## Proposed Changes
Adds TS 51.teams-messaging-extension-action


tracyboehrer:
This ran as expected in Teams except for one confusing element:  The README states to select "Create Card" from the Compos Box command list.  I didn't see that option.  I do see an "Action Messaging" option, which then displays "Create Card" (which does create/display a card).

stevkan:
@tracyboehrer, The short name in the `manifest.json` file is "Action Messaging Extension" which meets the 30 character length limit. However, the Compose Box command list appears to have a shorter limit (22 characters, I think) and is truncating the last word and trimming the name so only the first two words appear. This is only a problem in the command list. If the app is pinned to the send box, then the tooltip is visible with the full name.

I have a few options. I can:

1. Change the short name to "Create Card", but then the Compose Box command list will have a title of "Create Card" and subtitle of "Create Card". The latter of which would need to change to something different.
2. Shorten the short name to "Messaging Extension" as the tooltip will still read the full short name and the description will state its purpose (and update the README).
3. Do nothing (and update the README).

tracyboehrer:
It more that it doesn't match the Readme?
________________________________

stevkan:
I guess it depends on how you are reading it. After looking over the README again, I think the "Compose Box command list" is referring to this, which looks correct to me.

image

It's the "Messaging extensions" modal where I'm seeing the shortened name, but this isn't mentioned in the README.

image

PR #2636, opened 2020-07-30
Zoma/json keyword

### LINKED ISSUE #2510: Handle all standard JSON schema types and formats
We need to handle all of the built-in JSON schema types without any extra machinery. In particular we need to handle string, number, integer, boolean and array of them.  Most of this should be just doing the root template.  

In addition we need to handle standard formats from here: https://json-schema.org/understanding-json-schema/reference/string.html#format
In particular we should be able to hook-up date-time, time, date, email, regex and uri/iri.  I suspect we will need to change the generator to use the format name as the type rather than just string.  With this change we can make templates that hook-up email to the LUIS email, URI to URL and the date/time ones to datetimev2.

Currently we always generate an if condition for min/max on numbers.  Make the template generation smarter to check at generation time for min/max and only generate the check if present.  We should use the same structure for also checking the pattern regex that can be specified in JSON schema for strings.

We need to update the unittest.schema to include examples of all properties and we should have tests to make sure they are all properly handled.

[enhancement]

coldplaying42:
Fixes #2510 

## Proposed Changes
Support Json keyword maximum, minimum and pattern.


experimental/generation/generator/packages/library/src/dialogGenerator.ts
```diff
@@ -709,9 +709,10 @@ export async function generate(
 
         // Expand schema expressions
         let expanded = expandSchema(schema.schema, scope, '', false, true, feedback)
-
+      
         // Write final schema
         let body = stringify(expanded, (key: any, val: any) => (key === '$templates' || key === '$requires' || key === '$templateDirs' || key === '$examples') ? undefined : val)
+
         await generateFile(ppath.join(outPath, `${prefix}.json`), body, force, feedback)
```
chrimc62, line 717: Did you mean to change this?

experimental/generation/generator/packages/library/templates/generator.lg
```diff
@@ -104,6 +104,73 @@ ${'#'} ${property}Entity_getValueText(val)
...
+- ```
+{
+    "$schema": "${appSchema}",
+    "$kind": "Microsoft.OnAssignEntity",
+    "operation": "Add(${property})",
+    "entity": "${entity}",
+    "actions": [
+        {
+            "$kind": "Microsoft.IfCondition",
+            "condition":"not(isMatch( @${entity},  dialogClass.schema.properties['${property}'].pattern))",
```
chrimc62, line 117: >.pattern [](start = 95, length = 8)

Do you know if the regex language of isMatch and JSON Schema match?

experimental/generation/generator/packages/library/templates/generator.lg
```diff
@@ -104,6 +104,73 @@ ${'#'} ${property}Entity_getValueText(val)
...
+                    "value": "= @${entity}"          
+                }
+            ]
+        }
+    ]               
+    ${designerOpEntity('Add', entity)}     
+}
+```
+# addDialogWithNumberMaxMin(entity)
+- ```
```
chrimc62, line 141: Insert blank line.

experimental/generation/generator/packages/library/templates/generator.lg
```diff
@@ -104,6 +104,73 @@ ${'#'} ${property}Entity_getValueText(val)
...
+                    "property": "$${property}",
+                    "value": "= @${entity}"          
+                }
+            ]
+        }
+    ]               
+    ${designerOpEntity('Add', entity)}     
+}
+```
+# addDialogWithNumberMaxMin(entity)
```
chrimc62, line 140: >MaxMin [](start = 21, length = 6)

MinMax would be more traditional.

experimental/generation/generator/packages/library/templates/standard/date-timeProperty.lg.lg
```diff
@@ -0,0 +1,27 @@
...
+> Template for prompting the user for the '${property}' property.
+${'#'} ${property}_getPromptText
+- \${getStringPromptText('${property}')}
+```
+
+# ValueTemplate
+-```
+> Template for generating display text for the value of the '${property}' property.
+${'#'} ${property}_getValueText(val)
+- \${getValueText(val)}
```
chrimc62, line 26: Are these different than the normal ones?

experimental/generation/generator/packages/library/templates/standard/library-Assign.en-us.lg
```diff
@@ -8,6 +8,15 @@
...
 -
 
+# getDateValidationText
+- Please enter the correct date format.
+
+# getTimeValidationText
+- Please enter the correct time format.
+
+# getPatternValidationText(pattern)
+- The input should follow the pattern ${pattern}
```
chrimc62, line 18: >${pattern} [](start = 38, length = 10)

Put a period at the end.

experimental/generation/generator/packages/library/templates/standard/library.en-us.lg
```diff
@@ -12,7 +12,7 @@
 - ${template(`${property}_getPropertyNameText`)}
 
 # getPropertyValueText(property, val)
-- ${template(`${property}_getValueText`, val)}
+- ${replace(replace(string(template(`${property}_getValueText`, val)), '\r\n', ''), '"', '\'')}
```
chrimc62, line 15: >replace(replace(string(template(`${property}_getValueText`, val)), '\r\n', ''), '"', '\'')} [](start = 4, length = 91)

Why is this needed?

chrimc62:
>                     "value": "=@number"          

Will this automatically get coerced to a string or do you need to be explicit?

---
Refers to: experimental/generation/generator/packages/library/templates/standard/numberEntity-add-string.dialog.lg:37 in 8661ee8. [](commit_id = 8661ee8d6a350506ca9ace7ec487f0bf33a303c0, deletion_comment = False)

chrimc62:
Some template naming issues.
1) There should not be integerEntity-add* templates.  There is no integerEntity and they should be covered by the numberEntity-add* templates.  Naming pattern is <entity>-add-<propertyType>.
2) Instead of date-time use dateTime.

experimental/generation/generator/packages/library/templates/standard/date-timeProperty.lg.lg
coldplaying42: This is the same as ObjectProperty.lg.lg

coldplaying42:
@boydc2014 @GeoffCoxMSFT Please help me review the code

PR #2638, opened 2020-07-30
Fixed #2637: Directory separator char no longer hard-coded

### LINKED ISSUE #2637: .NET sample no. 15 (attachments) doesn't work on macOS
## Sample information
1. Sample type: \samples\
3. Sample name: Handling attachments

## Describe the bug
.NET sample no. 15 doesn't work on macOS. Answering "1" or "3" to the welcome message, which should result in the display of an inline or uploaded attachment, respectively, results in the following error message:
```System.IO.FileNotFoundException: Could not find file '.../samples/csharp_dotnetcore/15.handling-attachments/Resources\architecture-resize.png'```
This is because the backward slash is hard-coded in `AttachmentsBot.cs` but macOS uses a forward slash.

## Screenshot
Screenshot 2020-07-30 at 15 27 08

bjoernross:
Fixes #2637 

## Proposed Changes
.NET sample no. 15 didn't work on macOS because the backward slash is hard-coded as a directory separator. This pull request fixes the issue on macOS. 

I haven't tried any other samples or systems.

### DIFF
--- a/samples/csharp_dotnetcore/15.handling-attachments/Bots/AttachmentsBot.cs
+++ b/samples/csharp_dotnetcore/15.handling-attachments/Bots/AttachmentsBot.cs
@@ -166,7 +166,7 @@ namespace Microsoft.BotBuilderSamples
         // Please consult the channel documentation for specifics.
         private static Attachment GetInlineAttachment()
         {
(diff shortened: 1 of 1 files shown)


cleemullins:
Eric, Dave, JJ, and myself reviewed and merged.

PR #2644, opened 2020-08-04
Fix bug where main dialog was not generated.

chrimc62:


experimental/generation/generator/packages/cli/package.json
```diff
@@ -15,6 +15,7 @@
     "@types/fs-extra": "^8.0.0",
     "@types/mocha": "^5.2.7",
     "@types/node": "^10.17.18",
+    "chai": "^4.2.0",
```
GeoffCoxMSFT, line 18: Why add this package if there isn't code added to use it?

PR #2627, opened 2020-07-28
TS Teams conversation bot sample

stevkan:
## Proposed Changes
Adds TS 57.teams-conversation-bot sample

PR #2673, opened 2020-08-17
Fix javascript Teams sample 57

EricDahlvang:
Clicking 'Message all Members', 'Who am I?', 'Delete Card' or 'Update Card' is resulting in an exception, due to no .Value property present on the CardActions.

(discovered here: https://github.com/microsoft/BotBuilder-Samples/pull/2627#issuecomment-674929439 )


EricDahlvang:
blocked: need to investigate, since Python and Dotnet samples also have this issue.

Is this a regression?  is .value now required, when it was not before?

EricDahlvang:
This turned out to be a Teams client issue.  .value should not be required

PR #2689, opened 2020-08-21
Update yeoman to 4.10.1

mrivera-ms:


generators/generator-botbuilder/generators/app/templates/core/package-with-tests.json.js
```diff
@@ -17,10 +17,10 @@
     },
     "dependencies": {
         "@microsoft/recognizers-text-data-types-timex-expression": "1.1.4",
```
EricDahlvang, line 19: should this be updated?  to 1.3.0 
https://www.npmjs.com/package/@microsoft/recognizers-text-data-types-timex-expression

stevengum: We're still using `"1.1.4"` in `botbuilder-ai` ([link](https://github.com/microsoft/botbuilder-js/blob/b05fb963cf063984817f978e8bc8a5678132da0a/libraries/botbuilder-ai/package.json#L25)). That doesn't preclude us from bumping the dependency here, but if we do we should probably bump the package's dependency.

PR #2696, opened 2020-08-25
Zoma/simplify condition v2

### LINKED ISSUE #2648: Simplify generated prompts
Today the generated prompts are componentized and make use of library functions.  This makes them harder to understand/modify.  To simplify this, we will move runtime logic into generation.  

Before:
```
> Template for prompting the user for the 'Bread' property.
# Bread_getPromptText
- ${getEnumPromptText('Bread')}
```

After:
```
> Template for prompting the user for the 'Bread' property.
# Bread_getPromptText
- ```
${getAutoHelpText()}
Which value do you want for bread?```
```

To do this we will add a promptWithHelp template to generator.lg and use that within the templates.  Similarly for min/max numbers we will figure out which prompt to use at generation rather than runtime.

[enhancement]

coldplaying42:
Fixes #2648 

## Proposed Changes

  - 
  -
  -


experimental/generation/generator/packages/library/templates/generator.lg
```diff
@@ -308,7 +311,28 @@ ${'#'} ${property}Entity_getValueText(val)
             ]
         }
     ]
-    ${designerMissing(property)}     
+    ${designerMissing(property)
+    }
```
chrimc62, line 315: Should be same line.

experimental/generation/generator/packages/library/templates/generator.lg
```diff
@@ -308,7 +311,28 @@ ${'#'} ${property}Entity_getValueText(val)
...
+    "priority": "=lib.missingPriority('${property}')",
+    "actions": [
+        {
+            "$kind": "Microsoft.SendActivity",
+            "activity": "\${getHelpText()}"
+        }
+    ]
+    
+    ${designerMissing(property)
+    }
```
chrimc62, line 335: Same line.

experimental/generation/generator/packages/library/templates/generator.lg
```diff
@@ -308,7 +311,28 @@ ${'#'} ${property}Entity_getValueText(val)
...
+```
+
+# helpDialog
+- ```
+{
+    "$schema": "${appSchema}",
+    "$kind": "Microsoft.OnEndOfActions",
+    "condition": "$retries > 0 && $lastIntent != 'Help'",
+    "runOnce": true,
+    "priority": "=lib.missingPriority('${property}')",
```
chrimc62, line 326: I suspect this should be a priority of 0, i.e. it should run before anything else.

chrimc62:
A few minor things, otherwise lets get this checked-in.  I want to integrate with the new LUIS model stuff which is working now.  Thanks for pushing through all the issues on this!  The next simplification we should probably do is to separate the current Ask trigger into two parts, one for missing and one for change.  That is connected to the new model I have though so I will do that.  What was the fix for the scope of exported functions issue?

chrimc62 (approved):
:shipit:

experimental/generation/generator/packages/library/templates/generator.lg
coldplaying42: I change priority to 0, but in that case, the help info will been shown many times if the wrong input is fed.

PR #2697, opened 2020-08-25
update packages to 4.10.0

johnataylor:
- update packages
- minor code change in use of channel adapters
- slack channel adapter events have changed from the preview version

### DIFF
--- a/samples/csharp_dotnetcore/53.teams-messaging-extensions-action-preview/TeamsMessagingExtensionsActionPreview.csproj
+++ b/samples/csharp_dotnetcore/53.teams-messaging-extensions-action-preview/TeamsMessagingExtensionsActionPreview.csproj
@@ -8,7 +8,7 @@
   <ItemGroup>
     <PackageReference Include="AdaptiveCards" Version="1.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.1" />
-    <PackageReference Include="Microsoft.Bot.Builder.Integration.AspNet.Core" Version="4.9.2" />
+    <PackageReference Include="Microsoft.Bot.Builder.Integration.AspNet.Core" Version="4.10.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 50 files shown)


samples/csharp_dotnetcore/53.teams-messaging-extensions-action-preview/TeamsMessagingExtensionsActionPreview.csproj
```diff
@@ -8,7 +8,7 @@
   <ItemGroup>
     <PackageReference Include="AdaptiveCards" Version="1.0.0" />
```
stevengum, line 9: Not part of your changes, but this version of AdaptiveCards differs from sample [54.teams-task-module](https://github.com/microsoft/BotBuilder-Samples/pull/2697/files#diff-8cfcd63486a3b3b65cb9bfd3926c0c21R9) (1.2.0).

Do you know if there's a reason they're not the same?

samples/csharp_dotnetcore/62.webex-adapter/Adapters/WebexAdapterWithErrorHandler.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.BotBuilderSamples.Adapters
     public class WebexAdapterWithErrorHandler : WebexAdapter
     {
         public WebexAdapterWithErrorHandler(IConfiguration configuration, ILogger<WebexAdapter> logger)
-            : base(configuration, logger)
+            : base(configuration, null, logger)
```
stevengum, line 14: This is the third time I've seen this change; was there a breaking change in the `Microsoft.Bot.Builder.Adapters.<channel-name>` adapters? Why didn't we create another `WebexAdapter()` overload that still took `IConfiguration` and `ILogger<T>` if the middle parameter is nullable?

johnataylor: are you suggesting we add that extra constructor in 4.11.0 ❓ in which case we should get an issue open.
i made these changes so this compiles - given **4.10.0** is out the door I don't see what else we can do at this point. the change isn't breaking in this case because the channel adapters were actually preview (yes technically this is version one of the channel adapters 😄)

PR #2719, opened 2020-09-03
Support outlining for .lg

### LINKED ISSUE #2711: Enhance adaptive tools to support outlining for .lg/.lu
It would be really nice in .lu/.lg/.qna files to have a way of doing outlining on complex files.  Normally in markdown you could use number of #, but I suspect for our purposes making using of the > comments would be better.  A normal comment would have one > and an "outlining" one would have two or more >.
 
So for example:
```
>> Dialog templates
# foo
- something
# blah
- else
 
> Simple comment
# woof
- more stuff
>>> Special dialog templates
# bark
- woof
```

I would be able to collapse ">> Dialog Templates" and see only that and if I showed immediate children would show.  
```
>> Dialog templates
# foo
# blah
# woof
>>> Special dialog templates
```
Each of those could be expanded/collapsed, so I could expand the foo template by clicking on '# foo'.  Markdown editors do something similar and show a chevron on each section of expand/collapse.

ShYuPe:
Fixes #2711 

Add foldingRange function for .lg file.

foldingRange

### DIFF
--- a/experimental/adaptive-tool/package.json
+++ b/experimental/adaptive-tool/package.json
@@ -220,7 +220,9 @@
         "build:server": "webpack --mode production --config ./server/lg/webpack.config.js && webpack --mode production --config ./server/lu/webpack.config.js",
         "clean": "rimraf client/out && rimraf server/out",
         "postinstall": "cd client && npm install && cd ../server && npm install && cd ..",
-        "test": "sh ./scripts/e2e.sh"
+        "test": "sh ./scripts/e2e.sh",
+		"compile": "tsc -b",
+		"watch": "tsc -b -w"
     },
     "devDependencies": {
         "@types/node": "^13.9.3",
--- a/experimental/adaptive-tool/package.json
+++ b/experimental/adaptive-tool/package.json
@@ -220,9 +220,7 @@
         "build:server": "webpack --mode production --config ./server/lg/webpack.config.js && webpack --mode production --config ./server/lu/webpack.config.js",
         "clean": "rimraf client/out && rimraf server/out",
         "postinstall": "cd client && npm install && cd ../server && npm install && cd ..",
-        "test": "sh ./scripts/e2e.sh",
-		"compile": "tsc -b",
-		"watch": "tsc -b -w"
+        "test": "sh ./scripts/e2e.sh"
     },
     "devDependencies": {
         "@types/node": "^13.9.3",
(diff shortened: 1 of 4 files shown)


experimental/adaptive-tool/server/lg/src/providers/foldingRange.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as util from '../util';
+
+export function foldingRange(params: FoldingRangeParams, documents: TextDocuments<TextDocument>) {
+	const document = documents.get(params.textDocument.uri)!;
+	const item: FoldingRange[] = [];
+	if (!util.isLgFile(path.basename(document.uri))) {
+		return item;
+	}
+	const lineCount = document.lineCount;
+	for (var i = 0;i < lineCount;i++) {
```
cosmicshuai, line 13: it will be better to use `let i = 0` in the for loop. Due the function scope.

experimental/adaptive-tool/server/lg/src/server.ts
```diff
@@ -199,7 +205,7 @@ connection.onDidChangeWatchedFiles(_change => {
 			}
 		}
 	});
-	connection.console.log('We received an file change event');
+	// connection.console.log('We received an file change event');
```
cosmicshuai, line 208: maybe we can just remove this line.

experimental/adaptive-tool/package.json
```diff
@@ -220,7 +220,9 @@
         "build:server": "webpack --mode production --config ./server/lg/webpack.config.js && webpack --mode production --config ./server/lu/webpack.config.js",
         "clean": "rimraf client/out && rimraf server/out",
         "postinstall": "cd client && npm install && cd ../server && npm install && cd ..",
-        "test": "sh ./scripts/e2e.sh"
+        "test": "sh ./scripts/e2e.sh",
+		"compile": "tsc -b",
```
cosmicshuai, line 224: are these two lines necessary?

experimental/adaptive-tool/server/lg/src/providers/foldingRange.ts
ShYuPe: fixed

experimental/adaptive-tool/server/lg/src/server.ts
ShYuPe: Sure. I also remove some other console

experimental/adaptive-tool/package.json
ShYuPe: just run build is enough, I have removed these two lines

boydc2014:
cc @chrimc62

boydc2014:
We should also do this for .lu and also .qna maybe

chrimc62 (approved):
:shipit:

boydc2014 (approved):

PR #2721, opened 2020-09-07
Support outlining for lu

### LINKED ISSUE #2711: Enhance adaptive tools to support outlining for .lg/.lu

ShYuPe:
Fixes #2711 

Add foldingRange function for .lg file.

foldingRangeLu

### DIFF
--- a/experimental/adaptive-tool/client/src/extension.ts
+++ b/experimental/adaptive-tool/client/src/extension.ts
@@ -62,7 +62,8 @@ function startLgClient(context: ExtensionContext) {
 				didChangeWorkspaceFolders: ((data, next) => {
 					next(data);
 				})
-			}
+			},
+			provideFoldingRanges: (document, context, token, next) => next(document, context, token)
 		},
 		diagnosticCollectionName: 'lg'
 	};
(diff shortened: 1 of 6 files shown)


cosmicshuai (approved):

chrimc62 (approved):
:shipit:

boydc2014 (approved):

PR #2723, opened 2020-09-08
Adjust keybinding functions

ShYuPe:
In this pr, I move the keybinding function of .lg from the server side to the client side. 

This move accelerate the on-Enter command.


chrimc62 (approved):
:shipit:

boydc2014 (approved):

PR #2748, opened 2020-09-17
Fixing file name case

sethunath:
Fixing file name case so that the app works without errors in linux


sgellock (approved):
Thank you for the PR!  We really appreciate it.

PR #2751, opened 2020-09-20
Corrected missing text property on msg object

stevkan:
## Proposed Changes
On lines 119 and 140 of UserProfileDialog.cs, the sample incorrectly passes the message object, `msg.ToString()`, as a response to a disclosed/non-disclosed age. Instead, it should pass the message's text value, `msg.Text.ToString()`.

Line 31 of UserProfileDialog.lg and UserProfileDialog.fr-fr.lg incorrectly checks for a `null` value when the user selects "No" as a response. It should check for a `-1` which is the value that is actually returned.

### DIFF
--- a/samples/csharp_dotnetcore/language-generation/05.a.multi-turn-prompt-with-language-fallback/Dialogs/UserProfileDialog.cs
+++ b/samples/csharp_dotnetcore/language-generation/05.a.multi-turn-prompt-with-language-fallback/Dialogs/UserProfileDialog.cs
@@ -76,7 +76,7 @@ namespace Microsoft.BotBuilderSamples
 
             // We can send messages to the user at any point in the WaterfallStep.
             await stepContext.Context.SendActivityAsync(ActivityFactory.FromObject(_lgGenerator.Generate("AckName", new {
-                Result = stepContext.Result
+                stepContext.Result
             }, stepContext.Context.Activity.Locale)), cancellationToken);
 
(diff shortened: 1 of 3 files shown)


samples/csharp_dotnetcore/language-generation/05.a.multi-turn-prompt-with-language-fallback/Dialogs/UserProfileDialog.cs
```diff
@@ -116,7 +116,7 @@ private async Task<DialogTurnResult> ConfirmStepAsync(WaterfallStepContext stepC
             }, stepContext.Context.Activity.Locale));
 
             // We can send messages to the user at any point in the WaterfallStep.
-            await stepContext.Context.SendActivityAsync(ActivityFactory.FromObject(msg.ToString()), cancellationToken);
+            await stepContext.Context.SendActivityAsync(ActivityFactory.FromObject(msg.Text.ToString()), cancellationToken);
```
EricDahlvang, line 119: Something still looks off here.  There is a call to ActivityFactory.FromObject above, online 113.  The 2nd call to ActivityFactory seems redundant.  

Should this just be sending out the msg as it is?

stevkan: You are right. I hadn't looked at some of the surrounding code. This happens in both instances. I tested and sending just the constructed `msg` is enough.

EricDahlvang (approved):

PR #2853, opened 2020-10-20
add string luis entity schema

### LINKED ISSUE #2804: Ensure schemas exist for all LUIS entity types
We need to ensure that there are .schema for all LUIS entity types.  So far, we have only schematized ones that are formats or are objects.  We need to add schema for personname, phone number, keyphrase, percentage.   We should also change schema processing so that when $entities are in the item of an array they are promoted to the property level.  That way the elements of an array can point to one of the component schemas.

coldplaying42:
Fixes #2804 

## Proposed Changes

  - 
  -
  -


chrimc62 (approved):
:shipit:

coldplaying42:
@boydc2014 Please help me approve it.

Issue #2856, opened 2020-10-20
Samples reorganization


mrivera-ms:
- [x] 1. Move the **samples/csharp_dotnetcore/language-generation** folder content to https://github.com/microsoft/BotBuilder-Samples/tree/main/experimental/language-generation

- [ ] 	2. Move the **samples/csharp_dotnetcore/adaptive-dialog** folder content to a **csharp_dotnetcore** folder inside https://github.com/microsoft/BotBuilder-Samples/tree/main/experimental/adaptive-dialog.

- [x] 3. Move the adaptive-dialog sample **samples/csharp_dotnetcore/adaptive-dialog/01.multi-turn-prompt to 105.multi-turn-prompt** inside https://github.com/microsoft/BotBuilder-Samples/tree/main/samples/csharp_dotnetcore

- [x] 	4. Move the adaptive-dialog sample **adaptive-dialogs/05.interruptions-bot** to **190.interruptions-bot in samples/csharp_dotnetcore**.

- [ ] 	5. Create a new adaptive-dialog sample **143.complex-dialog** in  samples/csharp_dotnetcore.

mrivera-ms:
@EricDahlvang, @johnataylor I am closing this issue for now. If we need to make these changes later, we can reopen it.

PR #2858, opened 2020-10-20
Move 'samples/csharp_dotnetcore/language-generation' to 'experimental'

mrivera-ms:
Step 1 of #2856


EricDahlvang (approved):

chrimc62:
Not sure what you are looking for in a review.  This does indeed move the project to experimental.  Not sure what composer would do with this--but the adaptive tools vs code tools should work fine with the .lg files and this looks to be mainly code first so that is probably OK.

chrimc62 (approved):
:shipit:

PR #2859, opened 2020-10-21
Move adaptive-dialog/05.interruptions-bot to 190.interruptions-bot

mrivera-ms:
Step 4. of #2856


EricDahlvang (changes requested):
Can we move this to 109, rather than 190.

mrivera-ms:
@johnataylor could you please review?

mrivera-ms:
> Can we move this to 109, rather than 190.

Done.

Issue #2891, opened 2020-11-04
Orchestrator JS sample #1 has missing required parameter for luis:build command in README


tsuwandy:
Need to add --botName param for bf luis:build section, like the following:

> bf luis:build --in CognitiveModels --authoringKey <YOUR-KEY> --botName <YOUR-BOT-NAME>

PR #2892, opened 2020-11-04
Orchestrator samples update 

tsuwandy:
Fixes #
#2891 

## Proposed Changes
-Updated missing botname param in luis:build command section JS dispatch sample readme
-Removed Any CPU from C# .sln since only X64 build will work


CezaryMarcjan (approved):

scheyal (approved):
Also document the role of orchestrator.json so that it doesn't confuse users who will see duplicate entries in there and appsettings.json. They may also just copy the values (and I"m not sure if the full paths would work).

PR #2958, opened 2020-11-25
Zoma/lu file

### LINKED ISSUE #2957: type: string generates assets with ${property} in file names and content
To repro use a simple schema like this:
{
    "$schema": "http://json-schema.org/draft-07/schema",
    "type": "object",
    "properties": {
        "name": {
            "type": "string"
        }
    },
    "required": [
        "name"
    ]
}

If you generate when doing the stringEntity it will use genericEntity which has ${property} in it.

coldplaying42:
Fixes #2957

Missing awaits caused unstable merges.
Added help for PropertyToChange.
Fixed schema bug for string.


chrimc62 (approved):
:shipit:

PR #2961, opened 2020-11-27
resolve string entity

### LINKED ISSUE #2957: type: string generates assets with ${property} in file names and content

coldplaying42:
Fixes #2957

## Proposed Changes

  - 
  -
  -


chrimc62:
Would be good to fix the generic ones as well, but this works for now.

chrimc62 (approved):
:shipit:

PR #2970, opened 2020-12-04
Zoma/cross train1

### LINKED ISSUE #2893: Ensure adding QnA properly adds cross train recognizer
In order for a dialog to be loaded as part of tests all dependent dialogs must be present which is a problem when we rely on build for adding recognizers.  This was fixed for standard schema by generating the multi-lingual recognizer as a template and removing that responsibility from luconfig.json.  I suspect that we need to do the same thing for the cross-train recognizer when QnA is added.  We also need some tests that include professional schema.

coldplaying42:
Fixes #2893 

## Proposed Changes

  - 
  -
  -


experimental/generation/generator/packages/library/templates/professional/config.json.lg
```diff
@@ -11,7 +11,7 @@ ${join(foreach(locales, locale, localeTemplate(locale)), ','+EOL())}
 ```
 
 # localeTemplate(locale)
--```    "./${knowledgeDir()}${locale}/${prefix}.${locale}.lu": {
+-```    "${prefix}.${locale}": {
```
chrimc62, line 14: >${prefix} [](start = 9, length = 9)

Why did you change this?  For composer this is where the results need to be.

experimental/generation/generator/packages/library/templates/standard/build.cmd.lg
```diff
@@ -45,8 +45,8 @@ ${join(foreach(locales, locale, copyForLocale(locale)) , EOL())}
 
 # copyForLocale(locale)
 -```
-call copy .\cross-trained\\${prefix}.${locale}.lu .\\${locale}\ /y
-call copy .\cross-trained\\${prefix}.${locale}.qna .\\${locale}\ /y
+call copy .\cross-trained\\${prefix}.${locale}.lu .\language-understanding\\${locale}\ /y
+call copy .\cross-trained\\${prefix}.${locale}.qna .\knowledge-base\\${locale}\ /y
 ```
```
chrimc62, line 50: Wouldn't this be better in the template filename rather than done here?

chrimc62 (changes requested):
:clock1:

experimental/generation/generator/packages/library/templates/standard/build.cmd.lg
chrimc62, line 50: This would be better in the template filename rather than relying on this script which will not be run by composer.

experimental/generation/generator/packages/library/templates/professional/config.json.lg
```diff
@@ -11,7 +11,7 @@ ${join(foreach(locales, locale, localeTemplate(locale)), ','+EOL())}
...
             "rootDialog": true,
 		    "triggers": {"":""}
```
chrimc62, line 16: >	    "triggers": {"":""} [](start = 1, length = 24)

"out": ${knowledgeDir}

experimental/generation/generator/packages/library/templates/professional/config.json.lg
coldplaying42: I asked Fei, he said in Composer, the config file for cross train will be generated automatically.  Could we keep the change?

chrimc62 (approved):
:shipit:

PR #51, opened 2017-01-12
[Node-Samples] Update to botbuilder 3.5.1

pcostantini:

PR #2990, opened 2020-12-16
PR for New sample Codes

Trinetra-MSFT:
justinTimeInstallationBot
InlineImageDownload- Bot
SubscribeToBotConversationEvents

Fixes #

## Proposed Changes
1. These are new sample codes for JustinTimeInstallation for Bot, InlineImageDownload and SubscribeToConversationEvents


clearab:
@Wajeed-msft & @Trinetra-MSFT - I love that you guys want to improve these samples, and I really don't want to discourage that at all. But can you please check with us before adding completely new ones? Any new sample here creates an ongoing maintenance cost, and we need to make sure everything rationalizes correctly. Can you schedule some time for us to talk through this in January, and we can get some Issues created to track this work and make sure it get's prioritized (and the changes get ported into all the other languages)?

Based on my quick look through this:

- the events sample should be combined with the existing sample #57
- the just in time install should be added to sample #51 
  - get members is an expensive call to test if you're installed, should probably use getTeamDetails instead (fairly certain this call works in all contexts, despite the name)
  - The nested catch is a little awkward, hopefully there's a way to clean that up?
- Image download should probably also be added to sample #57, with the logic for getting the image abstracted into its own function. The try/catch here is also a little off, is there a better way?

@johnataylor & @EricDahlvang does the feedback above look correct to y'all? I'm a little worried about sample #57 getting overly large.

EricDahlvang:
'inline image download' might be better combined with '56.teams-file-upload' and changed to '56.teams-files' or something.?  I think we demonstrate this same principal in the '15.handling-attachments' sample.

clearab:
Looking at 15, I think that is probably the right one to combine with.

Trinetra-MSFT:
Thanks @clearab , @EricDahlvang , I will look into your review comments and try to accommodate these samples in existing ones.

clearab:
I'm going to go ahead and close this PR - I'll schedule some time for us to talk in early January to agree on how this functionality should get added.

Trinetra-MSFT:
@clearab , I worked on your review comments and merged code in existing samples, We will discuss these changes in Jan and understand your inputs on these so we can raise PR again. Thanks

PR #3015, opened 2021-01-08
Create go.yml

zinwayz:
Fixes #

## Proposed Changes

  - 
  -
  -


ceciliaavila (commented):
Thanks for this PR, but we are not planning to add GO to the workflow.
@tracyboehrer, we should close this PR.

PR #3019, opened 2021-01-12
Zoma/array object

### LINKED ISSUE #3018: Missing templates for schema $ref arrays
We are missing the templates to handle our schema references are array elements.  For example:
```
  "foo": {
      "type": "array",
      "items": {
        "$ref": "template:dimension.schema"
      }
    }
```
Does not generate because of "Missing template objectArray, Missing template generic-objectArray".
 We should handle this for all our built-in LUIS types.

coldplaying42:
Fixes #3018 

## Proposed Changes

  - 
  -
  -


chrimc62:
Could you have done this with a single generic object template rather than have one for every single one?

chrimc62 (changes requested):
:clock1:

PR #3104, opened 2021-03-01
Zoma/new lu merging

### LINKED ISSUE #3066: Improve merging for generated LU files
We need to improve merging for LU files.  To do this:
1) We should import a file custom.lu where people can put new utterances.
2) To delete an utterance in a generated file they should comment out the utterance.  To modify they would add the new label to custom utterances after deleting.

When we regenerate we would:
1) Overwrite everything in the generated LU files unless an unlabeled utterance corresponds to a commented out unlabeled utterance in which case we should leave it commented out with the new labels.  We should also delete any commented out utterance that is no longer being generated.
2) In user.lu we should rename any changed labels and remove any utterance that contains a label that doesn't exist anymore or there is an enum entity and the underlying phrase is no longer in the enumeration.

This provides us with a model where people can customize an LU model and the generation can continue to be updated.

coldplaying42:
Fixes #3066 

## Proposed Changes

  - 
  -
  -


experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -747,3 +707,30 @@ async function parseSchemas(schemaName: string, oldPath: string, newPath: string
...
+ * @param newPath Path to the folder of the new asset.
+ */
+async function getSynonyms(schemaName: string, newPath: string): Promise<Map<string, Set<string>>> {
+    let template = await fs.readFile(ppath.join(newPath, schemaName + '.json'), 'utf8')
+    let newObj = JSON.parse(template)
+    let propertyValueSynonyms = new Map<string, Set<string>>()
+
+    for (let property in newObj['properties']) {
+        let examples = newObj['properties'][property]['$examples']
+        if (examples !== undefined && examples[''][`${property}Value`] !== undefined) {
```
chrimc62, line 723: There are 4 possible sources of $examples in order.
1. property/$examples/locale/entity
2. property/$examples/""/entity
3. $examples/locale/entity
4. $examples/""/entity

experimental/generation/generator/packages/library/templates/standard/custom.lu.lg
```diff
@@ -0,0 +1,12 @@
+[generator.lg](../generator.lg)
+
+# filename
+- ${understandingDir()}${locale}/form/${prefix}-custom.${locale}.lu
+
+# template
+-```
+> Add custom utterances
+> Label with: Add(), Remove(), Help(), Clear(), Show() and Help()
```
chrimc62, line 9: Maybe add all of the labels and the hierarchy, i.e. Values: BreadValue, CheeseValue
Properties: BreadProperty(BreadValue), Length(dimension, number)
Operations: Add(PROPERTY), ...

experimental/generation/generator/packages/library/test/merge.test.ts
```diff
@@ -264,18 +264,13 @@ describe('dialog:generate --merge library', async function () {
             assertRemovedProperty(comparison, 'Toppings', errors)
             assertRemovedProperty(comparison, 'Sauces', errors)
             await assertContains('language-generation/en-us/BreadValue/sandwichMerge-BreadValue.en-us.lg', /black/, errors)
-            await assertContains('language-understanding/en-us/Bread/sandwichMerge-Bread-BreadValue.en-us.lu', /black/, errors)
             await assertMissing('language-generation/en-us/BreadValue/sandwichMerge-BreadValue.en-us.lg', /white/, errors)
```
chrimc62, line 267: Maybe you want to keep these and adjust the modified schema to include the $examples with black in it.

chrimc62 (changes requested):
:clock1:

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,150 +292,191 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
-                for (let enumSyn of removedEnumValue) {
-                    if (intentBodyStatement.match(enumSyn)) {
-                        matching = true
-                        break
-                    }
-                }
-                if (!matching) {
-                    intentResult.push(intentBodyStatement)
+const valuePattern = /\{@([a-zA-Z0-9]+)Value=([a-zA-Z0-9\s]+)\}/g
+
```
chrimc62, line 424: This works for now, but a more robust way of handling this would be to pick up the first $entities for the enum property.

experimental/generation/generator/packages/library/src/mergeAssets.ts
chrimc62: This is still wrong.  You need to be checking the specific locale you are generating, i.e. for foo.en-us.lu you look for a locale in the property then the default locale, then the same in the global $examples.

---
In reply to: [587057402](https://github.com/microsoft/BotBuilder-Samples/pull/3104#discussion_r587057402) [](ancestors = 587057402)

experimental/generation/generator/packages/library/src/mergeAssets.ts
coldplaying42: Do you mean matching the enum entity directly?

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -739,3 +780,38 @@ async function parseSchemas(schemaName: string, oldPath: string, newPath: string
...
+    let propertyValueSynonyms = new Map<string, Set<string>>()
+
+    for (let property in newObj['properties']) {
+        let examples = newObj['properties'][property]['$examples']
+        if (examples !== undefined) {
+            let valueExamples
+            if (examples[''][`${property}Value`] !== undefined) {
+                valueExamples = examples[''][`${property}Value`]
+            }else if (examples[locale][`${property}Value`] !== undefined) {
+                valueExamples = examples[locale][`${property}Value`]
```
chrimc62, line 801: These are backwards.  Locale specific first then fallback.

chrimc62: Also you are not checking the global $examples.

---
In reply to: [588570829](https://github.com/microsoft/BotBuilder-Samples/pull/3104#discussion_r588570829) [](ancestors = 588570829)

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -739,3 +780,38 @@ async function parseSchemas(schemaName: string, oldPath: string, newPath: string
...
+            }else{
+                continue
```
chrimc62, line 803: I'm not a fan of continue in most situations--it is a non-local exit that is less obvious than alternatives.  In this case just wrap the next block in an if (valueExamples).

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,191 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
-                let matching = false
-                for (const enumSyn of removedEnumValue) {
-                    if (intentBodyStatement.match(enumSyn)) {
-                        matching = true
-                        break
-                    }
-                }
-                if (!matching) {
-                    intentResult.push(intentBodyStatement)
+const valuePattern = /\{@([a-zA-Z0-9]+)Value=([a-zA-Z0-9\s]+)\}/g
```
chrimc62, line 423: >\{@([a-zA-Z0-9]+)Value=([a-zA-Z0-9\s]+)\} [](start = 22, length = 41)

To be more robust you should use balancing expressions like this: (?<open>{@)([^=]+)=(.*)(?<-open>}) $1 will be the key and $2 the value.

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -739,3 +780,58 @@ async function parseSchemas(schemaName: string, oldPath: string, newPath: string
...
+ */
+ async function getSynonyms(schemaName: string, newPath: string, locale: string): Promise<Map<string, Set<string>>> {
+    let template = await fs.readFile(ppath.join(newPath, schemaName + '.json'), 'utf8')
+    let newObj = JSON.parse(template)
+    let propertyValueSynonyms = new Map<string, Set<string>>()
+
+    if (newObj['$examples'] !== undefined) {
+        let globalExamples
+        if (locale in newObj['$examples']) {
+            globalExamples = newObj['$examples'][locale]
```
chrimc62, line 797: Make use of null operator and properties should be before global, i.e. 
const propExamples = newObj['properties'][property]['$examples']
const globalExamples = newObj['$examples']
const examples = propExamples?.[locale] ?? propExamples?.[''] ?? globalExamples?.[locale] ?? globalExamples?.['']
const valueExamples = examples?.[`${property}Value`]

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,189 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
-            // add  new enum entity in the new  lu file 
-            for (const [key, values] of enumValueMap) {
-                if (!enumSet.has(key)) {
-                    resultStatements.push('    - ' + key + ' :')
-                    for (const newStatement of values) {
-                        resultStatements.push(newStatement)
-                    }
-                }
-            }
+const valuePattern = /(?<open>{@)(?<lable>[^=]+)=((?<value>[a-zA-Z]*)|($1))/g
```
chrimc62, line 354: >lable [](start = 36, length = 5)

label

chrimc62, line 354: >[a-zA-Z] [](start = 59, length = 8)

This is not enough.  What if there is punctuation?

chrimc62, line 354: >$1 [](start = 71, length = 2)

What does this mean?

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,189 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
-                if (!keepEnumValue.has(enumSyn)) {
-                    removedEnumValue.add(enumSyn)
-                }
-            }
+/**
+ * @description: Generate the common pattern of the utterances.
+ * @param line Line of the lu file.
+ */
+async function generatePatternUtterance(line: string): Promise<string> {
+    line = line.replace('-', '').trim()
```
chrimc62, line 410: >line.replace('-', '').trim() [](start = 11, length = 28)

Too aggressive.  What would happen if I had the string foo-blah in my text?  I think I led you astray because I've been thinking C# rather than node.  

line.replace(/{@?[^=]+=|}|^>+\s*-?\s*|^-\s*|^\s*|\s*$/gm, '') will generate the utterance and you should do this any line that starts with '>'. It will also handle the - > and spaces at start/'end.

experimental/generation/generator/packages/library/src/mergeAssets.ts
coldplaying42: I find that typescript does not support balancing group.
(?<open>{@)(?<label>[^=]+)=(?<value>.*)(?<-open>}) is illegal for typescript.

coldplaying42: I do not find the regex pattern for typescript which is similar to balancing group supported by .Net

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,181 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
+ * @param filename Name of the lu file.
+ * @param newFileList List of old file paths.
+ * @param delUtteranceSet Set of deleted utterance patterns.
+ */
+async function getDeletedUtteranceSet(filename: string, oldFileList: string[], delUtteranceSet: Set<string>): Promise<void> {
+    const filePath = oldFileList.filter(file => file.match(filename))[0]
+    const text = await fs.readFile(filePath, 'utf8')
+    const lines = text.split(os.EOL)
+    for (let line of lines) {
+        if (line.startsWith('>') && line.match(propertyPattern)) {
```
chrimc62, line 369: >(line.startsWith('>') && line.match(propertyPattern)) [](start = 11, length = 53)

Use a regex of /^>\s*-/m Otherwise you can't comment out an utterance like "- cancel". Looking for *Property or *Value is not enough since labels might not be present.

experimental/generation/generator/packages/library/templates/standard/custom.lu.lg
chrimc62: We still need more documentation here.

---
In reply to: [587060618](https://github.com/microsoft/BotBuilder-Samples/pull/3104#discussion_r587060618) [](ancestors = 587060618)

experimental/generation/generator/packages/library/src/mergeAssets.ts
coldplaying42: Is there any common pattern which could check the existence of example utterance?

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,182 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
+            resultRefs.push(ref)
+            await updateCustomLUFile(schemaName, oldPath, newPath, oldFileList, mergedPath, locale, feedback)
+        }
+    }
 
-        for (const newEntitySection of newEntitySections) {
-            if (newEntitySection.Name !== oldListEntitySection.Name) {
-                continue
-            }
+    let val = resultRefs.join(os.EOL)
```
chrimc62, line 344: >let [](start = 4, length = 3)

const

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,182 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
-                            }
-                        } else {
-                            enumValueMap.set(enumEntity, temp)
-                            i = j - 1
-                            break
-                        }
-                    }
-                }
-            }
+    let patternIndex = oldText.search(GeneratorPattern)
```
chrimc62, line 346: >let [](start = 4, length = 3)

const

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,182 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
+ * @description: Get the set of deleted utterance patterns.
+ * @param filename Name of the lu file.
+ * @param newFileList List of old file paths.
+ * @param delUtteranceSet Set of deleted utterance patterns.
+ */
+async function getDeletedUtteranceSet(filename: string, oldFileList: string[], delUtteranceSet: Set<string>): Promise<void> {
+    const filePath = oldFileList.filter(file => file.match(filename))[0]
+    const text = await fs.readFile(filePath, 'utf8')
+    const lines = text.split(os.EOL)
+    for (let line of lines) {
```
chrimc62, line 369: >let [](start = 9, length = 3)

const

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,182 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
+        if (line.match(commentOutPattern) && line.match(propertyPattern)) {
```
chrimc62, line 370: >&& line.match(propertyPattern) [](start = 42, length = 30)

Remove this test--otherwise people cannot comment out utterances without labels.

experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -294,148 +292,182 @@ async function mergeRootFile(schemaName: string, oldPath: string, oldFileList: s
...
+        if (line.match(propertyPattern)) {
+            const newLine = await replaceLine(line, propertyValueSynonyms)
+            resultLines.push(newLine)
+
+        } else {
+            resultLines.push(line)
+        }
+    }
+
+    let val = resultLines.join(os.EOL)
```
chrimc62, line 443: >let [](start = 4, length = 3)

const

chrimc62 (approved):
:shipit:

chrimc62:
Address the remaining open items. In particular don't filter on property pattern and beef up the descriptions in the custom file.

experimental/generation/generator/packages/library/src/mergeAssets.ts
coldplaying42: I think we'd better to use some pattern to constrain which utterance users could comment out, or else, if the user can comment out the following utterance:
@ ml Add() usesFeature AddPhrases
    -@ ml CheeseProperty usesFeature CheeseProperty
        - @ CheeseValue CheeseValue
The lu file might not be correctly trained.

chrimc62: If they comment out those parts the file is broken already and we can't prevent them.  There is no pattern we can do for this:
> - cancel
Other than the > - at the start.

---
In reply to: [594935847](https://github.com/microsoft/BotBuilder-Samples/pull/3104#discussion_r594935847) [](ancestors = 594935847)

PR #3126, opened 2021-03-17
experimental cleanup

sgellock:
Removed outdated samples or samples that promote outdated patterns


carlosscastro (approved):
:shipit: 

@ckkashyap  fyi we're deleting one of your experimental samples. IF you have any concerns please let us know and we can bring it back to life if we have a plan on how to keep it up to date and clear message of how it helps customers. Thanks!

PR #3149, opened 2021-04-06
Change publisher from Microsoft to BotBuilder

Danieladu:
## Proposed Changes
Change the Publisher of adaptive vscode extension from "Microsoft" to "BotBuilder".
image


Danieladu:
Close it for publisher `Microsoft` is valid now.

PR #3150, opened 2021-04-08
Zoma/dynamic list

### LINKED ISSUE #3132: Support DynamicList templates as part of standard schema
LUIS recognizers support the ability to define dynamic lists that are passed in at runtime.  To do this your LUIS model needs to have a place holder entity which is statically populated at training time and dynamically populated at runtime.  The LUIS recognizer would look like:
"recognizer": {
            "$kind": "Microsoft.LuisRecognizer",
            "applicationId": "=settings.luis.DynamicLists_test_en_us_lu.appId",
            "endpoint": "=settings.luis.endpoint",
            "endpointKey": "=settings.luis.endpointKey",
            "dynamicLists": "=[dialog.PersonValue, dialog.RoomNumberValue]"
}
Then we would have an action template to fill in the runtime definition of dialog.person and hang that off of beginDialog.  It would have a SetProperty(dialog.person, LISTSTRUCTURE).  The default definition would be the same one generated for the training time, but this would give people a place to dynamically calculate the list to pass in.

Inside of BeginDialog is an action which looks like:
```
{$kind: Microsoft.SetProperty, property: "dialog.PersonValue", value: "{
                    "entity": "PersonValue",
                    "list": [
                        {
                            "canonicalForm": "Chris McConnell",
                            "synonyms": [
                                     "Chris",
                                     "McConnell"
                            ]
                        },
                        {
                            "canonicalForm": "Zongyang",
                            "synonyms": [
                            ]
                        }
                    ]
                },
```

To actually make this dynamic someone could call the graph API and compute this structure to set in memory that would then be passed into the LUIS recognizers.

This schema should be of type string with $examples that look like enum $examples.  

Person: {$ref: "dynamiclist.schema", $examples: {"": {PersonValue: { "Christopher McConnell": ["Chris", "Christopher"]}}}.  Then from this you generate a PersonValue LUIS list with the $examples and you use the same name when passing in a dynamic list.

You can view this as an open-ended enum (which is why its type is just string) where you can statically define training data and at runtime supply dynamically defined definitions.

1) Define dynamiclist.schema.   Type is string and may be able to reuse enum templates.
2) Extend recognizer template to add dynamiclists and the list is empty if there are no dynamic templates.
3) For each dynamic property add a SetProperty action to the begin dialog for setting the dynamic list in memory.

Examples from tests: `botbuilder-dotnet-main\tests\Microsoft.Bot.Builder.AI.LUIS.Tests\tests\LuisAdaptiveRecognizerTests\DynamicLists.test.dialog`

coldplaying42:
Fixes #3132 

## Proposed Changes

  - 
  -
  -


experimental/generation/generator/packages/library/templates/standard/dynamiclist.schema
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "type": "string",
+    "$generator": {
```
chrimc62, line 4: I thought you needed $dynamicList: true here as well?

experimental/generation/generator/packages/library/templates/standard/form-BeginDialog.dialog.lg
```diff
@@ -13,18 +13,61 @@
...
-                }
-            ]
-        }
+        ${setActions()}
     ]
 }
 ```
+
+# setActions
+- IF: ${!empty(where(schema.properties, property, property.value.$dynamiclist == "true"))}
```
chrimc62, line 22: >"true" [](start = 81, length = 6)

This should be a boolean so you don't need the quotes around the true.

chrimc62, line 22: >IF: ${!empty(where(schema.properties, property, property.value.$dynamiclist == "true"))} [](start = 2, length = 88)

I would move this inside of setPropertiesForDynamicLists where if it is true you do ", <value>" otherwise an empty string.  That way you don't have to duplicate the body twice.

experimental/generation/generator/packages/library/templates/standard/form-BeginDialog.dialog.lg
```diff
@@ -13,18 +13,61 @@
...
+    ]
+    }
+}
+```
+
+# listElements(property)
+- IF: ${getProperty(property.value.$examples[""], concat(property.key, 'Value'))}
+    - ${join(foreach(indicesAndValues(getProperty(property.value.$examples[""], concat(property.key, 'Value'))), enum, `{"canonicalForm": "${enum.index}", "synonyms": [${synonymsForDynamicList(enum.value)}] }`), `, ${EOL()}`)}
+- ELSEIF: ${getProperty(property.value.$examples[locale], concat(property.key, 'Value'))}
+    - ${join(foreach(indicesAndValues(getProperty(property.value.$examples[locale], concat(property.key, 'Value'))), enum, `{"canonicalForm": "${enum.index}", "synonyms": [${synonymsForDynamicList(enum.value)}] }`), `, ${EOL()}`)}
```
chrimc62, line 71: Use propertyList(property) to get the examples--it has the proper fallback logic.

experimental/generation/generator/packages/library/templates/standard/luis.lu.dialog.lg
```diff
@@ -14,10 +14,13 @@
...
+    },
+    "dynamicLists": "=${addList()}"
 }
 ```
 
 # appName
 - ${prefix}_${replace(locale, '-', '_')}_lu
 
+# addList
+- [${join(foreach(where(schema.properties, property, property.value.$dynamiclist == "true"), result, `dialog.${result.key}Value`), `,`)}]
```
chrimc62, line 26: >"true" [](start = 84, length = 6)

No quotes should be a boolean.

chrimc62, line 26: >result.key [](start = 111, length = 10)

Is this correct?  I think you need an indicesAndValues around schema.properties, right?

chrimc62:
You need to add some tests.

chrimc62 (changes requested):
:clock1:

experimental/generation/generator/packages/library/templates/standard/dynamiclist.schema
chrimc62: You should also have a $template: "dynamicList" and a corresponding dynamicList.lg file which has the $templates you need to expand this.

---
In reply to: [609963431](https://github.com/microsoft/BotBuilder-Samples/pull/3150#discussion_r609963431) [](ancestors = 609963431)

coldplaying42: I think we do not need $template in dynamiclist.schema .

chrimc62: That is because you are making the type enum and not string.  

---
In reply to: [610484318](https://github.com/microsoft/BotBuilder-Samples/pull/3150#discussion_r610484318) [](ancestors = 610484318)

experimental/generation/generator/packages/library/templates/standard/dynamiclist.template
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "type": "string",
+    "$dynamiclist": true,
+    "$template": "dynamiclist",
```
chrimc62, line 5: I suspect we could get rid of $dynamiclist: true and look for the $template being dynamiclist.  Could you try that?

experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lg.lg
```diff
@@ -0,0 +1,49 @@
...
+-```
+# ${entity}_chooseEntity()
+> Choose between '${entity}' values
+- \${chooseEnumEntity()}```
+
+# propertyEntityValue
+-```
+# ${entity}_Value(entity) 
+> Display value for '${entity}'
+- \${choices(entity.value, '${enumName(entity)}', 'and')}```
```
chrimc62, line 25: We should just return the entity.value.  The entity is open-ended so what we generate statically is not really useful.

experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistProperty.en-us.lg.lg
```diff
@@ -0,0 +1,43 @@
...
+                        clearDefault())}
+${helpTemplates(repromptEnum(), helpEnum())}
+${showTemplate(showDefault())}
+${nameTemplate()}
+${enumValueTemplate()}
+```
+
+# repromptEnum()
+- IF: ${isArrayProperty(property)}
+    - ${propertyName()} includes multiple values from: \${possibleValuesList(`${property}`)}
```
chrimc62, line 25: >possibleValuesList [](start = 58, length = 18)

I don't think you can use this--the possible values is defined dynamically not at generation time.

experimental/generation/generator/packages/library/src/dialogGenerator.ts
```diff
@@ -995,7 +995,7 @@ export async function generate(
 
         if (!error) {
             // Write final schema
-            const body = stringify(expanded, (key: any, val: any) => (key === '$templates' || key === '$requires' || key === '$templateDirs' || key === '$examples' || key === '$template' || key === '$generator') ? undefined : val)
```
chrimc62, line 998: >key === '$examples'  [](start = 144, length = 20)

Why did you have to add this back?  $examples are just for the generation process so we should not need it in the final schema.

experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lg.lg
chrimc62: Instead of doing this, just call value(entit.value)

---
In reply to: [612049019](https://github.com/microsoft/BotBuilder-Samples/pull/3150#discussion_r612049019) [](ancestors = 612049019)

experimental/generation/generator/packages/library/templates/standard/en-us/form.en-us.lg
```diff
@@ -96,6 +96,17 @@ You can also ask for over all help, cancel the whole form or skip a step.```
 > Display value for possible values from an enum
 - [${join(possibleValues(property), ', ')}]
 
+# possibleValuesForDynamiclist(property)
+- IF: ${dialogClass.schema.properties[property].$examples['en-us'][propertyNameValue()]}
+- ${join(foreach(dialogClass.schema.properties[property].$examples['en-us'][propertyNameValue()], enum, `${enum.key}`), `, `)}
+- ELSEIF: ${dialogClass.schema.properties[property].$examples[''][propertyNameValue()]}
+- ${join(foreach(dialogClass.schema.properties[property].$examples[''][propertyNameValue()], enum, `${enum.key}`), `, `)}
+- ELSE:
+-
```
chrimc62, line 105: This should be coming from the memory location that defines the dynamic list--not from $examples.

experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lg.lg
```diff
@@ -0,0 +1,52 @@
...
+    - ${phrase(option)}```
+
+# cases
+- ${join(foreach(indicesAndValues(examples), enum, case(enum.index)), '')}
+
+# default
+-```
+  - DEFAULT: 
+    - no value
+```
```
chrimc62, line 51: You don't need this any more.  This only handles the "static: examples, not the dynamic ones.

experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lg.lg
```diff
@@ -0,0 +1,42 @@
...
+> Display value for '${entity}' enumeration value
+- SWITCH: \${value}```
+
+# case(option)
+-```
+  - CASE: \${'${option}'} 
+    - ${phrase(option)}```
+
+# cases
+- ${join(foreach(indicesAndValues(examples), enum, case(enum.index)), '')}
```
chrimc62, line 42: You still don;t need this.

experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lg.lg
```diff
@@ -0,0 +1,42 @@
...
+
+# filename
+- ${entityLGFilename()}
+
+# template
+-```
+${formImport('form.en-us.lg')}
+${formImport('form-Choose.en-us.lg')}
+${propertyEntityValue()}
+${propertyEnumValue()}
```
chrimc62, line 11: You don't need this.

PR #3179, opened 2021-04-19
Re-order migration folder. Create top migration, move migration V3V4 to it.

scheyal:
Re-ordering migration folder to allow for more migrations.
Details:
 
1.	Create a BotBuilder-Samples / Migration folder
2.	Move V3V4Migration underneath it (& keep the original V3V4 folder as a redirect … page moved, go there…)
3.	Create Dispatch folder and move sample 14 inside (Csharp/Node)
4.	Replace sample 14 above with Orchestrator style dispatch
 
Thus, the folder’s purpose would be:
1.	Host for legacy samples used in migration docs
a.	And specific migration samples / walkthroughs not in Ms-docs
2.	Migration documentation that doesn’t fit MS-docs (or staged similar to other doc efforts that stem from github and slowly transition to ms-docs
3.	Any other migration efforts such as Composer old RT to new RT, VA to Composer packages, V4 to V5.


daveta (approved):

tsuwandy (approved):

mrivera-ms (approved):
:shipit:

PR #3267, opened 2021-05-12
Zoma/response

### LINKED ISSUE #3249: Switch output templates to response templates
All of our templates should switch to response templates to better support using speech.  To do this we would introduce another level between what we have now and where dialogs show messages.  It would look like this:
```
# SendActivity_Greeting()
[Activity
    Text = ${SendActivity_Greeting_text()}
]

# SendActivity_Greeting_text()
- Welcome to your bot.
```

coldplaying42:
Fixes #3249

## Proposed Changes

  - 
  -
  -


chrimc62 (approved):
:shipit:

PR #3304, opened 2021-06-04
Fix singleton dialog merge

chrimc62:
Singleton dialog merge was not working properly in that it did not pick-up schema changes.  Fixed that by rewriting the dialog merge part and adding tests.
Also fixed some other minor issues like using match instead of endsWith which is both faster and more correct.

### DIFF
--- a/experimental/generation/generator/packages/library/src/mergeAssets.ts
+++ b/experimental/generation/generator/packages/library/src/mergeAssets.ts
@@ -6,7 +6,7 @@
 import * as fs from 'fs-extra';
 import * as ppath from 'path';
 import * as os from 'os';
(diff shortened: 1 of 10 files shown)


experimental/generation/generator/packages/library/src/mergeAssets.ts
```diff
@@ -419,23 +420,26 @@ async function generatePatternUtterance(line: string): Promise<string> {
...
+            changed = true
         } else {
             resultLines.push(line)
         }
     }
 
     const val = resultLines.join(os.EOL)
-    await writeToFile(oldPath, mergedPath, `${schemaName}-custom.${locale}.lu`, oldFileList, val, feedback)
+    await writeToFile(oldPath, mergedPath, filename, oldFileList, val, feedback)
 }
```
coldplaying42, line 443: using changed here

PR #3321, opened 2021-06-16
fix no examples issue for dynamiclist

### LINKED ISSUE #3318: Dynamiclist should populate the LU list with ** values if no examples
When we generate the LU list for a dynamiclist property it ends up empty if there are no $examples.  Despite this we generate values like `**fooValue1**`.  In this case those generated values should be populated in the list so that LUIS can recognize the list values when training.

coldplaying42:
Fixes #3318

## Proposed Changes

  - 
  -
  -

### DIFF
--- a/experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lu.lg
+++ b/experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lu.lg
@@ -1,4 +1,5 @@
 [standard.en-us.lg](standard.en-us.lg)
+[generator.lg](../../generator.lg)
 
 # filename
 - ${entityLUFilename()}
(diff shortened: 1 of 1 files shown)


experimental/generation/generator/packages/library/templates/standard/en-us/dynamiclistEntity.en-us.lu.lg
```diff
@@ -16,11 +16,15 @@ ${operations(`${entity}`, allSynonyms(), false)}
...
 
 # allSynonyms()
 - IF: ${examples}
     - ${flattenObject(examples)}
 - ELSE:
-    - 
+    - ${generatedEntityValues()}
+
+# generatedEntityValues
+ - ${[`**${entity}1**`, `**${entity}2**`, `**${entity}3**`]}
```
chrimc62, line 29: You should use entityValue(null) from the generator.  That way these are only defined in one place.

chrimc62 (changes requested):
:clock1:

PR #3343, opened 2021-06-30
reorg prompts

### LINKED ISSUE #3336: Push form prompts to a single file
We need to reorganize the form wide templates so that we have one file with all of the top-level prompts in it.  To do this:
1) Consolidate all of the *-form-* into a single file and call it <prefix>-form-extensions.lg
2) Seperate out any top-level prompt with a top-level activity and its text are in a single file <prefix>-form.lg.
3) Look at all of the top-level prompts to make sure they are as simple and obvious as possible.

coldplaying42:
Fixes #3336

## Proposed Changes

  - 
  -
  -


experimental/generation/generator/packages/library/templates/standard/standard.template
```diff
@@ -47,13 +47,12 @@
...
         "form-NoneIntent.lu",
         "form-triggerIntent.lu",
-        "form-Choose.lg",
-        "form-DisplayOperation.lg",
         "form-phrases.lu",
         "form-global.lu",
         "form-custom.lu",
-        "form-Show.lg",
-        "form.lg",
+        "form-DisplayOperation.lg",
```
chrimc62, line 53: This should be part of form-extensions.lg

chrimc62: Never mind.  I understand why it is in its own file.

chrimc62 (approved):
:shipit:

PR #3352, opened 2021-07-01
Python: Removed msrest dependency from #54

### LINKED ISSUE #3351: Python: 54.teams-task-module includes msrest dependency
And it doesn't specify a version.  This dependency isn't needed, and if the dev has a later version of msrest installed can cause a problem.

tracyboehrer:
Fixes #3351


tracyboehrer:
@dmvtech FYI

PR #3397, opened 2021-08-05
add templates for simple form dialog

### LINKED ISSUE #3388: PVA friendly prototype
I'd like to build a prototype where we simplify the LU model down to what PVA supports.  To do this:
1) We would create a new directory templates/simple with a new root schema simple.schema.
2) We would copy over and modify the existing templates to generate a simpler LU model without operators and maybe without any labels.

Changes would be:
* The only operation would be assign and we would drop all other utterances and the labels for assign.  We would also drop all triggers other than missing property, assign and the form wide ones.  
* Eventually we would only support the entity types that PVA supports, but for now we can use the LUIS ones to simplify testing.  I believe this includes enum, but we would have to get some details there.
*  The LU files would only have entity labels for property and value.  Eventually the entity labels would likely be dropped altogether.  (Because all of the entities are prebuilt or list and they correspond 1 to 1 to properties.)
* Triggers would only have the value field.  (It might be we are required to have a property field in which case we would have the one property for each entity type.) 
* Eventually we might be able to test this by using the Orchestrator recognizer as well, but we should hold off on that.

This would result in a bot that supports the event triggered model including richer prompts, but would run on the current PVA runtime.

coldplaying42:
Fixes #3388

## Proposed Changes

  - 
  -
  -

### DIFF
--- a/experimental/generation/generator/packages/library/templates/generator.lg
+++ b/experimental/generation/generator/packages/library/templates/generator.lg
@@ -97,14 +97,32 @@
         "generic-help.dialog"
     ]}
 
+# genericTemplatesSimple()
+- ${[
+        "genericEntity.lu",
+        "genericEntity.lg",
+        "genericProperty.lg",
+        "generic-missing.dialog",
+        "genericEntity-assign.dialog"
+    ]}
+
 # numericTemplates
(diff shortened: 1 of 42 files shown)


ceciliaavila:
This has been in draft for over 2 years. I think it can be closed @tracyboehrer.

PR #3538, opened 2021-10-20
Add SSO support to 46.teams-auth

### LINKED ISSUE #2935: Teams SSO samples
We need to create SSO samples for Teams. The sample should show how SSO works for both conversational bots as well as messaging extensions. This can be accomplished by one of the options below:

1. Updating existing TeamsAuth samples to use SSO (both bot & messaging extension)
2. Create new SSO sample that has both bot and messaging extension.
3. Create two new samples

The samples will need to ensure they deal with the duplication of SSO request issue that may occur if a user is logged into multiple teams clients.

There is an existing dotnet sample in the experimental folder that should be cleaned up as part of this effort: https://github.com/microsoft/BotBuilder-Samples/tree/main/experimental/teams-sso/csharp_dotnetcore .

The README should call out that SSO in Teams only works in 1-1 chats, and not group contexts.

EricDahlvang:
Fixes #2935


mrivera-ms (approved):
:shipit:

PR #3586, opened 2021-11-10
set headers for get request

chkap:
Fixes #

## Proposed Changes

  - Nodejs sample 56 failed when upload inline images, the root cause is that the http request header for authorization is not set.

## Testing

My local test verify the fix.


ryihan (approved):

ceciliaavila (commented):
Duplicated with 33611. @tracyboehrer, we should close this PR.

PR #3628, opened 2021-12-14
[#1848] Move python generators from Samples to "generators" - Delete

ceciliaavila:
Addresses [#1848](https://github.com/microsoft/botbuilder-python/issues/1848)

## Description
This PR deletes the `generators/python` folder as it was moved to the [BotBuilder-python](https://github.com/microsoft/botbuilder-python) repository.

### Detailed Changes
- Deleted generators/python folder with all its files and projects.

## Testing
These images show the before and after of the `generators` folder.
image


mrivera-ms (approved):
:shipit:

PR #3694, opened 2022-02-08
[Core Bot Sample] Add welcomeCard as Content so it will be accessible when published

lauren-mills:
Resolves ICM

This PR resolves an ICM for the following issue: 

Recently I was working on a sample where I have identified that the Cards folder associated with the core bot sample is not getting published causing error in the bot when a call to adaptive card is made.

We need help to fix the same

[TROUBLESHOOTING STEPS]

Tried deploying the sample :

https://github.com/microsoft/BotBuilder-Samples/tree/main/samples/csharp_dotnetcore/13.core-bot

Uploaded Image

When we see on kudu console the folder cards itself is not published

 
However when the below sample is tested the resources folder is getting published

Uploaded Image

### DIFF
--- a/samples/csharp_dotnetcore/13.core-bot/CoreBot.csproj
+++ b/samples/csharp_dotnetcore/13.core-bot/CoreBot.csproj
@@ -6,14 +6,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Content Remove="Cards\welcomeCard.json" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <EmbeddedResource Include="Cards\welcomeCard.json" />
-  </ItemGroup>
-
-  <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.1" />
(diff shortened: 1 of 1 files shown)


samples/csharp_dotnetcore/13.core-bot/CoreBot.csproj
```diff
@@ -5,14 +5,6 @@
...
   </PropertyGroup>
 
-  <ItemGroup>
-    <Content Remove="Cards\welcomeCard.json" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <EmbeddedResource Include="Cards\welcomeCard.json" />
-  </ItemGroup>
-
```
mrivera-ms, line 15: This will break the sample. The Cards folder is not published because it is consumed as an embedded resource. See here: https://github.com/microsoft/BotBuilder-Samples/blob/648a181d96184ca84204b1d82fa7ecfddc5adb19/samples/csharp_dotnetcore/13.core-bot/Bots/DialogAndWelcomeBot.cs#L45

PR #3761, opened 2022-05-27
Add luisApplicationId parameter to CICD pipeline

BruceHaley:
Fixes #minor

## Proposed Changes
That parameter deploys the luis__applicationId configuration setting to Azure, which is required for bots using LUIS to work.


BruceHaley:
Apparently luis__applicationId is NOT needed. Another test deploy without it, and the bot works. ??

PR #3784, opened 2022-06-28
Moved projects from 3.1 to net6.0

Philippe-Laval:
Fixes #

## Proposed Changes

  - 
  -
  -


tracyboehrer:
@Philippe-Laval Thanks for this!  There were some other PR's related to this already submitted which I just merged into our staging branch.  We won't be using this, but I did review this to check if there was anything missing from the earlier PR's.

PR #3799, opened 2022-07-25
Fix: TypeScript Sample Authentication Config

### LINKED ISSUE #3797: What is the correct method to create a CloudAdapter instance?
All Typescript samples follow the following pattern to instantiate a CloudAdapter where the mandatory parameter of type "Configuration" is passed a null value:

1. Create a "credentialsFactory with ConfigurationServiceClientCredentialFactory;
2. Creates a botFrameworkAuthentication with createBotFrameworkAuthenticationFromConfiguration passing as parameter "null" and the "credentialsFactory";
3. Create an "adapter" with CloudAdapter passing as a parameter the "botFrameworkAuthentication";

**What is the correct method to create a CloudAdapter instance?**

## Sample information
**1. Sample type: ** samples
**2. Sample language:**  typescript
**3. Sample name:** 
- 00.empty-bot
- 02.echo-bot
- 03.welcome-users
- 05.multi-turn-prompt
- 06.using-cards
- 16.proactive-messages
- 50.teams-messaging-extensions-search
- 51.teams-messaging-extensions-action
- 57.teams-conversation-bot
- 58.teams-start-new-thread-in-channel

## To Reproduce
1. Clone and install de solution according to instructions in the README
2. Configure tsconfig.json with compiler options strict, noImplicitOverride, emitDecoratorMetadata and experimentalDecorators as true
3. Build and run application

## Expected behavior
Normal build and started application

## Screenshots
image
image
image

anishprasad01:
Fixes #3797.

Existing samples showed an incorrect method of configuring authentication. While the Node samples have already been updated in #3792, the TS samples have not yet been updated. This PR rectifies that by modifying the call to pass in `process.env`. Thanks to @stevkan for help in correcting this issue.

## Proposed Changes

  - Updates config call in each sample to create a new `ConfigurationBotFrameworkAuthentication` instance and pass in `process.env` as `ConfigurationBotFrameworkAuthenticationOptions`.
  - Removes `credentialsFactory` from each sample.

## Testing

Built and ran each sample to verify functionality.


samples/typescript_nodejs/00.empty-bot/src/index.ts
```diff
@@ -8,7 +8,8 @@ import * as restify from 'restify';
 import {
     CloudAdapter,
     ConfigurationServiceClientCredentialFactory,
```
IllusionMH, line 10: Looks like `ConfigurationServiceClientCredentialFactory` no longer used in these files and can be removed as well

PR #3816, opened 2022-08-30
Update templates and pipelines for linux deploys

BruceHaley:
Fixes #minor

## Proposed Changes
Fix Python templates and pipelines for linux deploys.
Fix JS templates and pipelines for linux deploys.


BruceHaley:
Replaced by #3817

PR #3825, opened 2022-09-14
Remove NuGet 5.5.1 install from pipelines

BruceHaley:
Fixes #minor

## Proposed Changes
This lets the pipelines use a later version of NuGet.


BruceHaley:
Replaced by #3826

PR #3869, opened 2023-03-03
teams-file-upload-sample bug fix

singhk97:
Fixes #

https://github.com/OfficeDev/Microsoft-Teams-Samples/issues/632

## Proposed Changes

 
- image
When the `invoke_response` is not `None` but `invoke_response.body` is `None` a json response is returned by the server with the body as `null`. This causes the `Something went wrong...` message. The correct behavior would be to return an empty body. 
- when the upload file request returns 200 response code, we should not return `file upload fail`

### DIFF
--- a/archive/samples/python/56.teams-file-upload/app.py
+++ b/archive/samples/python/56.teams-file-upload/app.py
@@ -75,11 +75,14 @@ async def messages(req: Request) -> Response:
     invoke_response = await ADAPTER.process_activity(
         activity, auth_header, BOT.on_turn
     )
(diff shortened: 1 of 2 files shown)


tracyboehrer:
Thank you, but archived samples are unmaintained.

ChetanSharma-msft:
Sample is maintained here: https://github.com/OfficeDev/Microsoft-Teams-Samples/tree/main/samples/bot-file-upload/python
So, we will do changes and raise a PR.

PR #3911, opened 2023-10-05
Modified adaptiveCardsBot to send weather information based on city, …

achowe13:
…and added coresponding files weather.py and WeatherCompactCard.json.

Fixes #

## Proposed Changes

  - 
  -
  -


ceciliaavila (commented):
Hi @tracy, @achowe13, I reviewed this PR and I don't think we should change this sample's functionality. It's meant to be a basic example of how to include adaptive cards into a bot. Users can take this and adapt it as needed. We should keep it simple.

tracyboehrer:
@achowe13 Thanks for your PR.  I really appreciate the interest in improving the samples. We can't change this sample since it also matches the functionality for the other languages, and then any docs we have around it.

PR #3947, opened 2024-03-02
Update attachmentsBot.js to correct handle downloads from ms teams

drewB:
The code was using attachment.contentURL which is missing the auth token need to download when using ms teams.  As a result, the bot would get a 403 when trying to download.


tracyboehrer:
@drewB Normally all Teams functionality is in the Teams Samples repo.  But this looks like a safe and practical changes.  Thanks!

PR #3949, opened 2024-03-14
Fix Python broken proactive example to use CONFIG instead of SETTINGS

yellalena:
## Proposed Changes
The example didn't work because SETTINGS variable didn't exist in the current context (used to in previous versions of repo).
Updated the example to use CONFIG instead. Also changed app_id to uppercase APP_ID because that's what's defined in the config.


yellalena:
@tracyboehrer hi! this is a tiny followup to your latest commit in the python sample. please review :)

tracyboehrer (approved):

PR #3995, opened 2024-06-19
[DotNet] Add details to README in certificate samples

JhontSouth:
#minor

## Description
This PR adds more information to the README certificate samples to have an easier complete guide for SSL/TLS certificate configuration.

## Proposed Changes
  - Updated Readme of dotnet samples 84.bot-authentication-certificate and 85.bot-authentication-sni.


tracyboehrer:
@JhontSouth For SN+I, I believe we can leave out the steps for creating a certifcate outside KeyVault.  The reason being is that for MS folks, the cert needs to be from OneCert.  In fact, creating a self-signed cert in KeyVault for SN+I just won't work.  I will update the README once I have all the details.  Basically they have to onboard with OneCert, and then they can create SN+I compatible certificates.  Also, for SN+I, they can't upload the cert to the App Registration (I didn't notice if the READ still indicated that).  It will cause a different alert if they do (pinning certs is not allowed either).

tracyboehrer (approved):

PR #4045, opened 2025-04-15
Replace C# Bot with Python Bot in the 44 Python Sample.

53845714nF:
## Proposed Changes
Hello, I'm new to the Topic. I looked at the README from the 44 Python Sample. I think there is a Copy Error from the C# README. Furthermore, I replace the C# with Python.


53845714nF:
Why was that closed? It's still wrong in the documentation.

